PR #1, opened 2017-08-22
Initial setup

hamishwillee:


julianoes:
👍 cool, is there a link already?

hamishwillee:
@julianoes Yes

Initial cut of the doc set gitbook can be viewed here: https://hamishwillee.gitbooks.io/dronecore/content/en/

This is really pre-alpha - I don't yet know exactly what I'll be getting in terms of API ref, installable libraries (e.g. pip), or examples yet. Generally speaking though, these are the sections we'll need.

I've split the "Building" stuff into "Contribution" section. Assumption is that there will be libraries that users can just download for most cases and that actually building the core would be reserved for contributors. 

unfortunately gitbook.com is playing up for me - so while build time is 6 seconds locally, it is taking hours on the server. Upshot, if you look at https://hamishwillee.gitbooks.io/dronecore/content/en/ you will see yesterday's work.

julianoes:
Cool, I like it.

One challenge will be the APIs changing from version to version.

hamishwillee:
Yes. Versioning is supported, but I don't know how good it is.

PR #13, opened 2017-09-07
Update docker instructions to use bash

hamishwillee:
@julianoes This updates the instructions for building using Docker.  The main changes are that:
-  this downloads the image rather than building it in the main example. This is usually a lot faster. I still show how to build the image in a later section
- This opens a bash prompt. Main reason is that if you want to build the application then this is much easier in bash - and so are instructions: "just pretend you are on linux" :-)
- I do also show the same things you did, as sub sections.

### DIFF
--- a/en/contributing/build.md
+++ b/en/contributing/build.md
@@ -185,22 +185,70 @@ build/default/integration_tests_runner --gtest_filter="ExternalExampleHello"
 
 ## Building in Docker
 
-If you want to develop in docker, you can use the [Dockerfile](https://github.com/dronecore/DroneCore/blob/master/Dockerfile) based on Ubuntu 16.04.
+You can also build DroneCore on your host computer with a toolchain running in a [Docker](https://docs.docker.com/) container (this saves you from having to install or manage the toolchain directly). 
 
-Build the image:
-```
-docker build . -t dronecore
+The main steps are:
+
+1. Install Docker on your host computer. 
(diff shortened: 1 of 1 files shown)


hamishwillee:
So do I need to clean things - what if I don't? Is the implication that I just can't access the files?
Can I make myself the owner - ie pass current user into docker for the build?
If I run sudo make clean outside of docker, do I need anything special in my environment

julianoes:
> So do I need to clean things - what if I don't? Is the implication that I just can't access the files?

You just can't clean it as a user.

>  Can I make myself the owner - ie pass current user into docker for the build?

It's a bit complicated. There is the way how it is done for PX4 containers:
https://github.com/PX4/containers/blob/master/docker/px4-dev/scripts/entrypoint.sh
I want to try this some time but it's bit more work, so not immediately.

>  If I run sudo make clean outside of docker, do I need anything special in my environment

No.

en/contributing/build.md
```diff
@@ -185,23 +185,75 @@ build/default/integration_tests_runner --gtest_filter="ExternalExampleHello"
...
+   make default install
+   # Run code-style check
+   make fix_style
+   # Build the docs
+   make docs
+   # Clean the build
+   make clean
+   ```
+   
+> **Note** Files built in a Docker container are owned by root. In order to clean up the **build** and **install** folders you will need to either call `make clean` in the container or `sudo make clean` in the host computer.
```
hamishwillee: Thanks for clarification @julianoes . I added this note and merged.

julianoes: Great thanks.

PR #17, opened 2017-09-11
build: fix Dronecore -> DroneCore

julianoes:
This was wrong and would put you in an empty folder in docker.


julianoes:
@hamishwillee is your system case-sensitive? 😄

hamishwillee:
Ha ha. No, in case it isn't clear above, I screwed up.

PR #24, opened 2017-09-13
Add topic explaining how to add ports and connect to devices

hamishwillee:


en/guide/connections.md
```diff
@@ -0,0 +1,84 @@
...
+
+```cpp
+DroneCore dc;
+//... 
+//Use UUID named uuid to get the associated Device
+Device &device = dc.device(uuid);
+```
+
+> **Tip** If you know that there is only one connected device you can instead call `device()` without any UUID. This will 
+return the first device detected, or a null `Device` if none have been discovered.
```
hamishwillee: @julianoes 
- What are the properties of a Null device - ie can you access an Action for it? etc.
- Really, all I am after here is to understand  what is the easiest way to tell that you have a null device

hamishwillee:
@julianoes This is a pretty basic topic on adding connections. I hope to extend the monitoring ports stuff when I understand a bit more about serial ports and TCP ports.

en/guide/connections.md
hamishwillee: @julianoes Thanks. Merged. Can you tell me the best way to check whether a Device from device() is valid - ie is Null?

PR #25, opened 2017-09-14
Add topic explaining how to set up build files for Cpp apps

hamishwillee:
This explains 
- the requirements for a build system (essentially anything you like, as long as it can find the libs/headers for DroneCore AND it adds the needed thread library).
- The minimum setup if you want to use cmake, based on our example code.

It does feel a little ponderous (mostly because of the relative paths), but anyone wanting to build apps is going to have to go down this path, and I think having an explanation helps.

### DIFF
--- a/en/contributing/build.md
+++ b/en/contributing/build.md
@@ -254,23 +254,32 @@ You can also build the image yourself using the [Dockerfile](https://github.com/
    ```
 
 
-## Install Artifacts
+## Install DroneCore Headers and Static Library
 
-The build artifacts (headers and static library file) can be installed locally into the folder `./install/` using:
+In order to *use* DroneCore your C++ applications need to be able locate the DroneCore libraries and header files (see [Building C++ Apps](../guide/toolchain.md)). You can either install these to standard system locations or locally within the DroneCore directory.
+
+> **Note** When using the library **libdronecore.a**, you need to link to a thread library such as *pthread* on a POSIX system (pthread is not included in the static library because it is included in glibc).
+
+
+### System-wide Install
+
+Installing DroneCore system-wide means that the files you need to build your apps are always in the same place, and your build definition files can be simpler because they do not need to consider relative locations.
+
+> **Warning** System-wide install is not (yet) supported on Windows. If you're a Windows guru, we'd [love your help](../README.md#getting-help) to set this up.
+
+To install the files system-wide:
+
+```bash
+make clean  #REQUIRED!
+sudo INSTALL_PREFIX=/usr/local make default install
+```
+
+> **Note** System-wide install overwrites standard install locations. If you already have DroneCore installed through some other mechanism it will be replaced!
+
+### Local Install
+
+The DroneCore headers and static library can be installed locally into the folder **DroneCore/install/** using:
 
 ```
 make default install
 ```
-
-Or, to install the files system-wide, using:
-
-```
-INSTALL_PREFIX=/usr/local make default install
-```
-
-Note that when using the library **libdronecore.a**, you need to link to a thread library such as *pthread* on a POSIX system (pthread is not included in the static library because it is included in glibc).
-
-## Build Python Wrapper
-
-TBD.
-
--- a/en/contributing/build.md
+++ b/en/contributing/build.md
@@ -254,7 +254,7 @@ You can also build the image yourself using the [Dockerfile](https://github.com/
    ```
 
 
-## Install DroneCore Headers and Static Library
+## Install DroneCore Headers and Static Library {#install-artifacts}
 
 In order to *use* DroneCore your C++ applications need to be able locate the DroneCore libraries and header files (see [Building C++ Apps](../guide/toolchain.md)). You can either install these to standard system locations or locally within the DroneCore directory.
 
--- a/en/guide/toolchain.md
+++ b/en/guide/toolchain.md
@@ -6,7 +6,7 @@ DroneCore C++ apps are written in standard C++ (c++11) and can be built using yo
 
 DroneCore itself uses the [cmake](https://cmake.org/) build system, and we recommend that you do too. CMake is an open-source, cross-platform toolchain that allows you to build your examples on Mac OS, Linux and Windows using the same build file definition.
 
-Below we explain how to set up a minimal make file for your application.
+Below we explain how to set up a minimal build setup (**CMakeLists.txt**) file for your application.
 
 
 ## Build Definition File - CMakeLists.txt
--- a/en/guide/toolchain.md
+++ b/en/guide/toolchain.md
@@ -1,0 +1,121 @@
+# Building C++ DroneCore Apps
+
+DroneCore C++ apps are written in standard C++ (c++11) and can be built using your preferred build system, compiler and linker toolchain. The only requirements are:
+- The build system must be able to locate the DroneCore headers and libraries (installed as described [here](../contributing/build.html#install-artifacts)).
+- When using the library **libdronecore.a**, you need to link to a thread library such as *pthread* on a POSIX system (*pthread* is not included in the static library because it is included in *glibc*). 
+
+DroneCore itself uses [cmake](https://cmake.org/), and we recommend that you do too. CMake is an open-source, cross-platform toolchain that allows you to build your examples on Mac OS, Linux and Windows using the same build file definition.
+
+Below we explain how to set up a minimal make file for your application.
+
+> **Tip** You can also kick off your application by copying modifying our existing [example code](https://github.com/dronecore/DroneCore/blob/master/example/takeoff_land/). Note that if you move your new application outside of the example directory you will need to update the (relative) paths to the locally installed DroneCore library.
+
+
+## Build Definition File - CMakeLists.txt
+
+*Cmake* uses a definition file named **CMakeLists.txt** to define the project. This specifies the name of the project, compiler flags for different platforms and targets, where to find dependencies (libraries and header files), source files to build, and the name of the generated binary. **CMakeLists.txt** is typically stored in the root directory of your app project.
+
+The first two lines of the file should specify the minimum compatible version of *cmake* and *your* project's name. 
+```cmake
+cmake_minimum_required(VERSION 2.8.12)
+project(your_project_name)
+```
+
+Next add the define flags that are used in compilation on different platforms 
+(you can treat this as boilerplate - it enables relatively strict handling of warnings).
+```cmake
+if(NOT MSVC)
+    add_definitions("-std=c++11 -Wall -Wextra -Werror")
+else()
+    add_definitions("-std=c++11 -WX -W2")
+    set(platform_libs "Ws2_32.lib")
+endif()
+
+# Add DEBUG define for Debug target
+set(CMAKE_CXX_FLAGS_DEBUG "-DDEBUG")
+```
+
+Add the line below to find a thread library on Linux, Mac, and Windows. This is required in order to use DroneCore!
+```cmake
+find_package(Threads REQUIRED)
+```
+
+The `include_directories()` command tells *cmake* where to find locally installed header files (if DroneCore is installed globally this is not required). 
+
+```cmake
+include_directories(
+    # Not needed if installed system-wide
+    ${CMAKE_SOURCE_DIR}/../../install/include
+)
+```
+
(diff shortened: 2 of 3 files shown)


en/guide/toolchain.md
```diff
@@ -0,0 +1,121 @@
+# Building C++ DroneCore Apps
+
+DroneCore C++ apps are written in standard C++ (c++11) and can be built using your preferred build system, compiler and linker toolchain. The only requirements are:
+- The build system must be able to locate the DroneCore headers and libraries (installed as described [here](../contributing/build.html#install-artifacts)).
+- When using the library **libdronecore.a**, you need to link to a thread library such as *pthread* on a POSIX system (*pthread* is not included in the static library because it is included in *glibc*).
```
hamishwillee: Windows uses the .lib right? Does it need to link against a thread library?

en/guide/toolchain.md
```diff
@@ -0,0 +1,121 @@
...
+
+Then use `add_executable()` to specify the name for your generated executable and your source files. 
+```cmake
+add_executable(your_executable_name
+    your_source_file.cpp
+)
+```
+
+The final sections of the file defines the libraries to link against. 
+The first part is used if DroneCore is installed locally, and will have to be modified depending on the relative path from your code to the **DroneCore/install/** directory.
```
hamishwillee: @julianoes Section below here is my "guess" on what the file does. Can you please check this a bit more carefully.

en/guide/toolchain.md
```diff
@@ -0,0 +1,121 @@
...
+```cmake
+if(WINDOWS)
+    set(dronecore_lib "${CMAKE_SOURCE_DIR}/../../install/lib/dronecore.lib")
+else()
+    set(dronecore_lib "${CMAKE_SOURCE_DIR}/../../install/lib/libdronecore.a")
+endif()
+```
+
+The final section of the file defines the libraries to link against if DroneCore is installed globally.
```
hamishwillee: @julianoes Ditto! It is not clear to me what this does, I THINK it locates the headers using that   ${platform_libs} variable.

en/guide/toolchain.md
```diff
@@ -0,0 +1,121 @@
...
+## Building the App
+
+This section assumes that you have already [built and installed the DroneCore C++ Library](../contributing/build.md) and that your example has a **CMakeLists.txt** in its root directory. 
+
+All we need to do is create a build directory (to keep all our build files in one place) and run the build from there. 
+
+```bash 
+cd /path/to/your/example/  # Open a prompt and navigate to your example 
+mkdir build && cd build   # Create a build directory and navigate into it
+cmake --build .  # Build the current directory
```
hamishwillee: @julianoes This line was taken from the example docs and didn't work. I tried `cmake ..` and that worked fine. What is "correct"?

en/guide/toolchain.md
julianoes: Yes, on Windows you need a threading library as well.

It's easiest to use the cmake functionality for this, see:
https://github.com/dronecore/DroneCore/blob/d85344800381d85f1e62467630c2d25c4c7ecab6/example/takeoff_land/CMakeLists.txt#L15-L16

and then link to: ${CMAKE_THREAD_LIBS_INIT}

Oh, you have that explained later. I don't remember what the file is called in Windows, could it be:
```
pthreadVC2.lib
pthreadVCE2.lib
pthreadVSE2.lib
```

en/guide/toolchain.md
julianoes: I think you're right. Just the paths are strange. The reason is that we install it locally inside `install`, instead we should probably support the usual install into `/usr/local/`, at least for Linux/Mac.

I made an issue for this: https://github.com/dronecore/DroneCore/issues/70

en/guide/toolchain.md
julianoes: No, `platform_libs` is a variable containing all libraries that you need to link to given your platform (e.g. pthread).

en/guide/toolchain.md
julianoes: You mixed Linux and Windows cmake arguments there.

To first, run cmake do:
```
mkdir build && cd build
cmake ..
```

Then the build step on Linux/Mac is (inside this build folder):
```
make -j8
```

And for Windows:
```
cmake --build .
```

en/guide/toolchain.md
```diff
@@ -0,0 +1,121 @@
...
+  * [cmake_minimum_required()](https://cmake.org/cmake/help/latest/command/cmake_minimum_required.html) - Minimum compatible version of cmake
+  * [project()](https://cmake.org/cmake/help/latest/command/project.html) - Set a name for project (and optionally version, languages, etc.)
+  * [add_definitions()](https://cmake.org/cmake/help/latest/command/add_definitions.html) - Adds -D define flags to the compilation of source files
+  * [set()](https://cmake.org/cmake/help/latest/command/set.html) - Set a normal, cache, or environment variable to a given value
+  * [find_package()](https://cmake.org/cmake/help/latest/command/find_package.html) - Load settings for an external project.
+  * [include_directories()](https://cmake.org/cmake/help/latest/command/include_directories.html)
+  * [add_executable()](https://cmake.org/cmake/help/latest/command/add_executable.html) - Specify executable to be created and set of source files to use
+  * [target_link_libraries()](https://cmake.org/cmake/help/latest/command/target_link_libraries.html) - Specify libraries to use when linking
+  * [message()](https://cmake.org/cmake/help/v3.9/command/message.html) - Display a message to the user
```
julianoes: Here you link to 3.9, in the others to latest. For my taste, that's almost too many links. I would probably slim it down a bit.

hamishwillee: OK. I'll change everything to latest. I'll move the links into the body of the doc and leave just the main entry point links here. Thanks.

en/guide/toolchain.md
hamishwillee: I don't think you actually need the -j8 bit for Linux do you?

en/guide/toolchain.md
hamishwillee: Thanks. I am pretty sure based on your comments that this text and our example code are broken for use with system wide install of dronecore. I have been assuming it can magically pick up that install, but in fact nowhere do we "tell" cmake were it can find the system wide headers and binary. I don't see how it could guess except for "well known libraries"

Consider, the first bit sets the variable $dronecore_lib  to the local paths for the platform.
```cmake
if(WINDOWS)
    set(dronecore_lib "${CMAKE_SOURCE_DIR}/../../install/lib/dronecore.lib")
else()
    set(dronecore_lib "${CMAKE_SOURCE_DIR}/../../install/lib/libdronecore.a")
endif()
```
Then in this bit we link dronecore_lib variable (defined above), our platform_libs variable, and we initialise the thread library. Nowhere do we set the system libraries. The comments below are confusing - this stuff has nothing to do with system wide or dronecore.
```cmake
target_link_libraries(takeoff_and_land
    ${dronecore_lib}
    # If installed system-wide:
    # dronecore
    ${CMAKE_THREAD_LIBS_INIT}  # Link against thread library found using find_package() 
    ${platform_libs}  # Variable containing all libraries that to link against for this platform (e.g. pthread).
)
```
To work system wide cmake needs to know where system libraries are installed. 
I'd suggest that we probably should use an environment variable, but assuming hard coding, this part of the makefile should change to:
```cmake
if(WINDOWS)
    set(dronecore_lib "${CMAKE_SOURCE_DIR}/../../install/lib/dronecore.lib")
    #Using local path -not sure right thing to do.
else()
    set(dronecore_lib "/usr/local/lib/libdronecore.a")
endif()

# Bit below stays the same - ${dronecore_lib} now points to system libs. 

target_link_libraries(takeoff_and_land
    ${dronecore_lib}   #Droncore libs, as defined above
    ${CMAKE_THREAD_LIBS_INIT}  # Link against thread library found using find_package() 
    ${platform_libs}  # Variable containing all libraries that to link against for this platform (e.g. pthread).
)
```

julianoes: The thing is that the linker automatically looks at `/usr/lib/` and (for most distros) `/usr/local/lib/` for libraries. This means that if you have it installed system-wide using prefix `/usr/local`, the linking is fine with just `dronecore`.

hamishwillee: Ignore my comment here. You explained this well over slack :-)

en/guide/toolchain.md
julianoes: No it's just faster.

en/contributing/build.md
```diff
@@ -254,23 +254,32 @@ You can also build the image yourself using the [Dockerfile](https://github.com/
    ```
 
 
-## Install Artifacts
+## Install DroneCore Headers and Static Library
```
hamishwillee: @julianoes OK, this is updated instructions for installing DroneCore. I've more clearly defined what "install" means, tried to clarify that system install is preferred and the risks, added call for action for Windows, and noted that system wide install requires the `make clean`.

en/contributing/build.md
```diff
@@ -254,23 +254,32 @@ You can also build the image yourself using the [Dockerfile](https://github.com/
...
 
-Or, to install the files system-wide, using:
 
-```
-INSTALL_PREFIX=/usr/local make default install
+### System-wide Install
+
+Installing DroneCore system-wide means that the files you need to build your apps are always in the same place, and your build definition files can be simpler because they do not need to consider relative locations.
+
+> **Warning** System-wide install is not (yet) supported on Windows. If you're a Windows guru, we'd [love your help](../README.md#getting-help) to set this up.
```
julianoes: @maetugr you do Windows, don't you?

en/contributing/build.md
```diff
@@ -254,23 +254,32 @@ You can also build the image yourself using the [Dockerfile](https://github.com/
...
+
+To install the files system-wide:
+
+```bash
+make clean  #REQUIRED!
```
julianoes: 👍

en/guide/toolchain.md
```diff
@@ -0,0 +1,213 @@
+# Building C++ DroneCore Apps
```
hamishwillee: @julianoes OK, I know you approved the PR, but I've also updated this one to address the issues you raised - and if would be good if you could cast your eye over it. 

Main changes:
- The build section at the end corrected as you said
- Divided into system-wide installers and local installers. The system installer section is turnkey - just replace these fields. The local install adds the explanation - I know which I'd rather do!
- Unfortunately gitbook does not support source code line highlighting. I've requested that. 
- Reduced the Useful link section you were annoyed by.

I think this is quite clear.

en/guide/toolchain.md
```diff
@@ -0,0 +1,213 @@
...
+
+DroneCore C++ apps are written in standard C++ (c++11) and can be built using your preferred build system, compiler and linker toolchain. The only requirements are:
+- The build system must be able to locate the DroneCore headers and libraries (installed as described [here](../contributing/build.html#install-artifacts)).
+- When using DroneCore you need to link to a thread library (e.g. *pthread* on a POSIX system). 
+
+DroneCore itself uses the [cmake](https://cmake.org/) build system, and we recommend that you do too. CMake is an open-source, cross-platform toolchain that allows you to build your examples on Mac OS, Linux and Windows using the same build file definition.
+
+Below we explain how to set up a minimal make file for your application.
```
julianoes: I would write `a minimal build setup` for your application. I think it's either `Makefile` or `CMakeLists.txt` but never really `make file`.

hamishwillee:
Thanks @julianoes (I fixed "I would write a minimal build setup for your application. I think it's either Makefile or CMakeLists.txt but never really make file." and merged. Good catch - that was horrible.

en/contributing/build.md
MaEtUgR: Yes but I almost always tried to work with MinGW/msys and GCC not the official Visual Studio way.
But I see that by the time I try to get my notifications under control you already helped yourself.

hamishwillee: Well, we haven't solved native windows build yet. Not sure if MinGW is "good enough" solution. @julianoes ?

julianoes: Last time I tried MinGW I could not stand it.

hamishwillee: My take on this is that I'm very impressed by what @MaEtUgR has done with MinGW but I see it as a bit "non standard". The learning curve is a bit high if you want to configure or change anything. 
My preference would be to use VisualStudio (which I see as the standard). Either way, we need to make this turnkey.

PR #32, opened 2017-09-26
Update apiref

hamishwillee:


en/guide/device_version.md
```diff
@@ -0,0 +1,43 @@
+## Device Information
```
hamishwillee: @julianoes I buggered up my merge and picked this new doc up as well - prior to review. I think it is probably OK, but can you please scan to see it is OK?

en/guide/device_version.md
```diff
@@ -0,0 +1,43 @@
...
+
+> **Note** Not all version information will necessarily be relevant on all vehicles. Where this occurs the 
+hardware may return garbage values (for example, the simulator provides garbage values for the vendor 
+firmware semantic version).
+
+The code below shows how to query the UUID and version information and print it to the console:
+
+```cpp
+// Get Device to query (in this case the first detected vehicle)
+Device &device = dc.device();
```
julianoes: For this line to succeed, a device needs already be connected. Usually you need to wait > 1s for the discover after calling `add_udp_connection`. Not sure if you want to add that.

hamishwillee: @julianoes Thanks. Our examples don't have this wait - if it is needed we should add it. 

Actually I think that if we successfully found a device via register_on_discover() this should not be able/allowed to fail. If it can, that is a bug. Thoughts? 

I should however make it clear that this assumes a device has been found. Perhaps change comment to (?):
```
// Get Device to query (this code assumes we have already found at least one vehicle and we want to query the first detected vehicle).
```

julianoes: Yes that would be good.

hamishwillee: Thanks. Done in https://github.com/dronecore/docs/commit/b45e56641a7a60301f6388c150ba1213d9ad6513

PR #38, opened 2017-09-27
build: automatically cleanup docker container

julianoes:
The argument `--rm` automatically cleans up leftover docker containers
after you exit the docker container.

To cleanup old containers, you can also use:
```
docker system prune
```

### DIFF
--- a/en/contributing/build.md
+++ b/en/contributing/build.md
@@ -223,7 +223,7 @@ The main steps are:
    ```
 1. Enter the following command in your host's terminal:
    ```sh
-   docker run -it -v $HOME/<path-to-dronecore-repo>/DroneCore:/home/docker1000/src/DroneCore:rw dronecore/dronecore bash
+   docker run --rm -it -v $HOME/<path-to-dronecore-repo>/DroneCore:/home/docker1000/src/DroneCore:rw dronecore/dronecore bash
    ```
    > **Note** The `-v` flag maps a directory on your host (left side) to a path in
    the container (right side). Above you need to specify the left-side path to the DroneCore repository on your host. The container path must be set as above.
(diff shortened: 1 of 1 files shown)


julianoes:
> Is there any case that I might want to not clean up my docker container when I exit.? The only case I can think of perhaps is the first/bash case - say I want to come back and reuse it. Is this likely?

In the case where you re-attach to containers you know how it works and you'll just remove the `--rm`. For the rest of us that just want to build it using docker, I think it's a sensible default.

en/contributing/build.md
```diff
@@ -223,7 +223,7 @@ The main steps are:
    ```
 1. Enter the following command in your host's terminal:
    ```sh
-   docker run -it -v $HOME/<path-to-dronecore-repo>/DroneCore:/home/docker1000/src/DroneCore:rw dronecore/dronecore bash
+   docker run --rm -it -v $HOME/<path-to-dronecore-repo>/DroneCore:/home/docker1000/src/DroneCore:rw dronecore/dronecore bash
    ```
    > **Note** The `-v` flag maps a directory on your host (left side) to a path in
```
hamishwillee: How about we add a note about the --rm flag, as below:

```
> **Note** The `-v` flag maps a directory on your host (left side) to a path in the container (right side). 
> You need to specify the left-side path to the DroneCore repository on your host and the container path must be set as above. 
> The `--rm` automatically cleans up leftover docker containers after you exit the docker container.
```

hamishwillee:
@julianoes OK, sounds very reasonable. How about we add a note explaining --rm too (see comment inline). Then happy to merge.

en/contributing/build.md
julianoes: Done 9a1896c

hamishwillee:
Thanks. Merged.!

PR #40, opened 2017-10-10
Explain how to make release builds

hamishwillee:


en/contributing/build.md
```diff
@@ -82,8 +87,6 @@ In order to *use* DroneCore your C++ applications need to be able locate the Dro
 
 > **Tip** The example code assumes that DroneCore is installed locally. The example **CMakeLists.txt** file can easily be modified to find the system-wide installation.
 
-<span></span>
```
hamishwillee: FYI I removed comment because this is now covered in building apps.

hamishwillee:
@julianoes 
1. As discussed the release build stuff might not work until https://github.com/dronecore/DroneCore/issues/91 is fixed, but I figure worth getting in now. Do you think this is sufficient?
2. I do wonder if we should have a "toolchain reference" listing all our make commands and the allowed modifiers (like `BUILD_TYPE=Release`). Thoughts? What other modifiers are supported?
3. How do we build a release build for Windows?

julianoes:
1. Should be fixed but we need to add a release build to travis and collect the artefacts.
2. This should be documented where the other build instructions are, I think.
3. For Windows it's a bit different, check: https://github.com/dronecore/DroneCore/blob/develop/appveyor.yml#L57-L61

hamishwillee:
@julianoes Thanks for the info, I updated with release and install instructions for windows - please check:

> This should be documented where the other build instructions are, I think.

What do you mean? I have put this as a Note in the flow of building?
I could have separate heading for this info "Building for release vs debug" if that is what you are suggesting?

julianoes:
I think the way you have it is good.

hamishwillee:
Thanks @julianoes . I also changed Mac OS X to macOS everywhere. So there!

PR #42, opened 2017-10-18
Add takeoff/landing doc - also covers other action APIs

hamishwillee:
This adds documentation for the Action API. 
- It is named "Takeoff and Landing" as those are the main APIs, and I don't think that "Action" will necessarily be recognised by readers. 
- It refers to both sync and async APIs but only shows sync examples. IMO that makes sense as a) sync usage is much easier to understand b) will most often be used.
- I don't show examples in every case, because the APIs are consistent. So I say something like "Used in the normal way"

### DIFF
--- a/en/guide/taking_off_landing.md
+++ b/en/guide/taking_off_landing.md
@@ -1,0 +1,220 @@
+# Take-off and Landing (and Other Actions)
+
+DroneCore provides the [Action](../api_reference/classdronecore_1_1_action.md) class for commanding the vehicle to arm, takeoff, land, return home and land, and disarm. 
+
+Most of the methods have both synchronous and asynchronous versions. The methods send commands to a vehicle, and return/complete with the vehicle's response. It is important to understand that a successful response indicates whether or not the vehicle intends to act on the command, not that it has finished the action (e.g. arming, landing, taking off etc.).
+
+> **Note** The implication is that you may need to monitor for completion of actions!
+
+<span></span>
+> **Note** All the code fragments below refer to the object `device`, which is a [connected device/vehicle](guide/connections.md).
+
+
+## Taking Off
+
+The recommended way to take off using DroneCore (and PX4) is to use either of the [takeoff()](../api_reference/classdronecore_1_1_action.md#classdronecore_1_1_action_1ae159c78b1a4056137187dc6e6d878539) or [takeoff_async()](../api_reference/classdronecore_1_1_action.md#classdronecore_1_1_action_1a2aec10a2b14f5e82f05edc6e2feac83e) methods. If a takeoff command is accepted the vehicle will change to the [Takeoff mode](https://docs.px4.io/en/flight_modes/takeoff.html), fly to the takeoff altitude, and then hover (in takeoff mode) until another instruction is received. 
+
+> **Note** PX4/DroneCore also provides other ways to take off:
+> - A copter or VTOL will take off automatically if a mission is started (fixed-wing will not). 
+> - You can also take off by manually driving the vehicle using the offboard API. 
+
+The vehicle will only take off once *armed*, and can only arm once it is "healthy" (has been calibrated, the home position has been set, and there is a high-enough quality GPS lock). After it starts flying, code needs to check that takeoff is complete before sending additional instructions.
+
+
+### Health Check
+
+The code fragment below shows very simple code to synchronously poll for health status (using [Telemetry:health_all_ok()](../api_reference/classdronecore_1_1_telemetry.md#classdronecore_1_1_telemetry_1a630c91d8067e4084c4f303513a0aeb29)) prior to arming.
+
+```cpp
+// Wait until health is OK and vehicle is ready to arm
+while (device.telemetry().health_all_ok() != true) {
+    std::cout << "Vehicle not ready to arm ..." << std::endl;
+    std::this_thread::sleep_for(std::chrono::seconds(1));
+}
+```
+
+The code fragment below performs the same task, but additionally exits the app if calibration is required reports what conditions are still required before the vehicle is "healthy":
+
+```cpp
+// Exit if calibration is required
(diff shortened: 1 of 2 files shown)


en/guide/taking_off_landing.md
```diff
@@ -0,0 +1,220 @@
...
+while (device.telemetry().health_all_ok() != true) {
+    std::cout << "Vehicle not ready to arm ..." << std::endl;
+    std::this_thread::sleep_for(std::chrono::seconds(1));
+}
+```
+
+The code fragment below performs the same task, but additionally exits the app if calibration is required reports what conditions are still required before the vehicle is "healthy":
+
+```cpp
+// Exit if calibration is required
```
hamishwillee: I haven't tested this part of the code because I can't get SITL to fail the calibration.

en/guide/taking_off_landing.md
```diff
@@ -0,0 +1,220 @@
...
+
+```cpp
+// Wait until health is OK and vehicle is ready to arm
+while (device.telemetry().health_all_ok() != true) {
+    std::cout << "Vehicle not ready to arm ..." << std::endl;
+    std::this_thread::sleep_for(std::chrono::seconds(1));
+}
+```
+
+The code fragment below performs the same task, but additionally exits the app if calibration is required reports what conditions are still required before the vehicle is "healthy":
```
hamishwillee: @julianoes  IMO This approach is much better than the default because it exits early in the case where vehicle cannot recover (requires calibration). Should we use this in examples?

en/guide/taking_off_landing.md
```diff
@@ -0,0 +1,220 @@
...
+        std::cout << ERROR_CONSOLE_TEXT << "  - Local position estimate." << NORMAL_CONSOLE_TEXT << std::endl;
+    }
+    if (!current_health.home_position_ok) {
+        std::cout << ERROR_CONSOLE_TEXT << "  - Home position to be set." << NORMAL_CONSOLE_TEXT << std::endl;
+    }
+    std::this_thread::sleep_for(std::chrono::seconds(1));
+}
+```
+
+> **Note** Vehicle health can also be checked using [Telemetry:health_all_ok_async()](../api_reference/classdronecore_1_1_telemetry.md#classdronecore_1_1_telemetry_1a83b384cd04b2ed17db805cfad8bafab5).
```
hamishwillee: Note again, not showing async usage, just pointing to API.

en/guide/taking_off_landing.md
```diff
@@ -0,0 +1,220 @@
...
+
+if (arm_result != Action::Result::SUCCESS) {
+    std::cout << "Arming failed:" 
+      << Action::result_str(arm_result) 
+      <<  std::endl;
+    return 1; //Exit if arming fails
+}
+```
+
+Use [Telemetry::armed()](../api_reference/classdronecore_1_1_telemetry.md#classdronecore_1_1_telemetry_1a0ca7da7922c22509ce6d55d4ad19bcf7) to check that the vehicle has armed before calling the takeoff command.
```
hamishwillee: Originally I just had a timer here. Then I tried using FlightMode::Ready - but during startup my SITL showed FlightMode::Hold, so decided just checking `armed` is more robust. 

@julianoes - use this in example code?

en/guide/taking_off_landing.md
```diff
@@ -0,0 +1,220 @@
...
+        takeoff_result) << std::endl;
+    return 1;
+}
+```
+
+If the command succeeds the vehicle will takeoff, and hover at the takeoff altitude. Code should wait until takeoff has completed before sending the next instruction. Unfortunately there is no specific indicator to inform code that takeoff is complete. 
+
+> **Note** One alternative is to simply wait for enough time that the vehicle *should* have reached the takeoff altitude.
+
+The code below checks for takeoff completion by polling the current altitude until the target altitude is reached:
```
hamishwillee: @julianoes When PX4 moves to HOLD after takeoff we can revisit this.

en/guide/taking_off_landing.md
```diff
@@ -0,0 +1,220 @@
...
+while (device.telemetry().armed()) {
+    std::this_thread::sleep_for(std::chrono::seconds(1));
+}
+std::cout << "Disarmed, exiting." << std::endl;
+```
+
+
+## Return/RTL
+
+[Return mode](https://docs.px4.io/en/flight_modes/rtl.html) (also known as "Return to Launch", "Return to Land", "Return to Home") flies the vehicle back to the home position and may also land the vehicle (depending on vehicle configuration). This mode is invoked from DroneCore using the [return_to_launch()](../api_reference/classdronecore_1_1_action.md#classdronecore_1_1_action_1a9af73101ce850e37cf7259dcdeda2eb9) or [return_to_launch_async()](../api_reference/classdronecore_1_1_action.md#classdronecore_1_1_action_1aa1253c356c7628d329dfa98d78eb39ee) methods.
```
hamishwillee: There is a big discussion on return modes going on now. For this purpose lets just ignore that :-)

en/guide/taking_off_landing.md
hamishwillee: @julianoes OK, I will remove arm code.

Should we update example code to include the calibration checks? I'd argue yes, because it is best practice.  What about also adding the more verbose status info - should that also be added to example code.

en/guide/taking_off_landing.md
hamishwillee: Yes. There are a couple more possible modes and variations on modes in the works - rally points (return to a safe point) and follow original path back to safety.

en/guide/taking_off_landing.md
hamishwillee: Yes it is. I'd welcome your proposal for a more concise approach, but this will do for now.

en/guide/taking_off_landing.md
julianoes: Yes, that would be nice.

PR #56, opened 2017-11-03
Plugin, SDK and Testing

hamishwillee:
This is a first attempt to provide additional guidance on testing, plugins, and working with SDKs. 
- New doc on writing plugins. 
- New doc on writing SDKs
  - attempts to just cover differences with writing plugins
  - Pulls out info on building SDKs from build doc
- Update testing to have some information explaining how to write tests and the facilities provided by dronecore.
- Try make it clear that integration test should be written first, and examples last.


en/contributing/build.md
```diff
@@ -293,3 +243,11 @@ You can also build the image yourself using the [Dockerfile](https://github.com/
    docker run --rm -it -v $HOME/<path-to-dronecore-repo>/DroneCore:/home/docker1000/src/DroneCore:rw my_image bash
    ```
 
+## Build SDK/External Plugins {#sdk_plugins}
```
hamishwillee: @julianoes FYI, the bulk of the section on building SDK/plugins is now in the sdk doc. This is a very much reduced version I only included here for "completeness of the build options".

en/contributing/documentation.md
```diff
@@ -51,18 +51,22 @@ For setup information see: [Gitbook toolchain](https://toolchain.gitbook.com/set
 
 ## API Reference
 
-The C++ source code is annotated using comments using [Doxygen](https://www.stack.nl/~dimitri/doxygen/manual/index.html) syntax. You can extract the documentation on macOS or Linux when you [build the library](../contributing/build.md) using the command: 
+The C++ source code is annotated using comments using [Doxygen](https://www.stack.nl/~dimitri/doxygen/manual/index.html) syntax. You can extract the documentation to markdown files (one per class) when you [build the library](../contributing/build.md).
```
hamishwillee: @julianoes This is just a clarification/improvement of how our API Ref is generated. You don't need to review this doc.

en/contributing/plugins.md
```diff
@@ -0,0 +1,102 @@
+# Writing Plugins
```
hamishwillee: @julianoes OK, new doc on writing "standard" plugins. I'm still not completely happy with it. What it tries to capture is:
- Plugins all have same location - subfolder of /plugins
- Plugins all have same structure - with implementation etc. This structure is shown by the external example
- If both above are true for a plugin it will be auto added to DroneCore at build time. 
- Plugins are also defined for SDKs 
- Integration tests should be created for plugins
- Example code should be created too, but after tests.

en/contributing/plugins.md
```diff
@@ -0,0 +1,102 @@
...
+        ├── example.cpp
+        ├── example.h
+        ├── example_impl.cpp
+        ├── example_impl.h
+        └── example_impl_test.cpp  ##optional
+```
+
+## Create a Plugin
+
+To create a new C++ plugin, copy the [example](https://github.com/dronecore/DroneCore/tree/master/external_example/plugins/example/) plugin from the "external_example" into the DroneCore **plugins** directory.
```
hamishwillee: @julianoes Are they better off copying this external_example for core plugins or one of the existing plugins (e.g. "action"). Why?
The CMakeLists.txt look similar but not identical, which makes me think that copying an existing standard plugin would be better?

en/contributing/plugins.md
```diff
@@ -0,0 +1,102 @@
...
+Modify the plugin as needed and update its [CMakeLists.txt](https://github.com/dronecore/DroneCore/blob/master/external_example/plugins/example/CMakeLists.txt) as appropriate:
+* Add additional libraries using the variable `additional_libs`:
+  ```
+  set(additional_libs "library_name" PARENT_SCOPE)
+  ```
+* Add required includes with `additional_includes`:
+  ```
+  set(additional_includes "include_dir" PARENT_SCOPE)
+  ```
+<!-- * You can also add tests with `unittest_source_files`, as discussed in the following section. -->
```
hamishwillee: This section was copied by me from the "SDK" plugin docs. I have commented out this line on `unittest_source_files` because it is not clear if it is used. I THINK it is, if present. Can you please confirm?

IF this stuff is exactly the same for both standard/external plugins I might just document it here and link from SDK docs.

en/contributing/plugins.md
```diff
@@ -0,0 +1,102 @@
...
+
+The [standard plugins](https://github.com/dronecore/DroneCore/tree/master/plugins) can be reviewed for guidance on
```
hamishwillee: @julianoes This is all I have on the actual code for a plugin. Is it worth saying more - ie showing some code to intercept a particular MAVLink message and process it?

en/contributing/plugins.md
```diff
@@ -0,0 +1,102 @@
...
+how to send and process MAVLink messages.
+
+## Test Code
```
hamishwillee: @julianoes Do you think it is OK to link to the tests doc here, or would you prefer everything about testing plugins to be in the writing plugins doc. I am in two minds. If we move here, then the testing doc would just be about running tests.

en/contributing/plugins.md
```diff
@@ -0,0 +1,102 @@
...
+> Do not write example code until the plugin has been accepted!
+
+A simple example should be written that demonstrates basic usage of its API by 3rd parties. 
+The example need not cover all functionality, but should demonstrate enough that developers 
+can see how it is used and how the example might be extended.
+
+Where possible examples should demonstrate realistic use cases such that the code
+can usefully be copied and reused by external developers.
+
+The plugin example should be documented in markdown following the same pattern as the existing [examples](../examples/README.md).
```
hamishwillee: Might move this down into Documentation section.

en/contributing/test.md
```diff
@@ -77,3 +87,138 @@ To run all telemetry tests:
 ```
 make && build/default/integration_tests_runner --gtest_filter="SitlTest.Telemetry*"
 ```
+
+## Writing Tests {#writing_tests}
```
hamishwillee: @julianoes So this section added to provide information about writing the tests. I am very tempted to move this to the plugins section, but it is about testing so it is relevant here. Thoughts?

en/contributing/test.md
```diff
@@ -77,3 +87,138 @@ To run all telemetry tests:
...
+{
+    ASSERT_TRUE(true);
+}
+
+} // namespace dronecore
+```
+
+#### Adding Unit Tests
+
+In order to run a unit test its file needs to be added to the DroneCore unit test program (`unit_tests_runner`). For the core plugins this is done in the main DroneCore [DroneCore/CMakeLists.txt](https://github.com/dronecore/DroneCore/blob/master/CMakeLists.txt#L187) file:
```
hamishwillee: @julianoes I think this is wrong - I was confused by the fact that the only unit tests we have are in the core. I think that for a plugin they should just define their unit tests in unit_tests_src in the same way as sdk plugins. Of course that doesn't explain what  `${plugin_unittest_source_files}` is for?

en/contributing/test.md
```diff
@@ -77,3 +87,138 @@ To run all telemetry tests:
...
+
+```cpp
+#include <iostream>
+#include <unistd.h>
+#include "dronecore.h"
+#include "integration_test_helper.h"
+
+using namespace dronecore;
+
+TEST_F(SitlTest, ExternalExampleHello)
```
hamishwillee: @julianoes I think it is useful to explain that this is needed for our integration tests.

en/contributing/test.md
```diff
@@ -77,3 +87,138 @@ To run all telemetry tests:
...
+    std::vector<uint64_t> uuids = dc.device_uuids();
+    EXPECT_EQ(uuids.size(), 1);
+
+    // Apparently it can say hello.
+    dc.device().example().say_hello();
+}
+```
+
+#### Adding Integration Tests
+
```
hamishwillee: @julianoes I think this can be simplifed in that looking at this external and internal plugins do the same thing to add integration tests. It is just that the CMakeLists.txt is in different places. So I'd tidy this to make that clear.

en/examples/README.md
```diff
@@ -1,5 +1,7 @@
 # Examples
 
+> **Tip** Information about *writing* example code is covered in the [Contributing > Writing Plugins](../contributing/plugins.md) (*plugin developers* should initially create [integration tests](../contributing/test.md#integration-tests) rather than examples for new code).
```
hamishwillee: @julianoes This was my first effort to make people aware that they should not write and example before integration tests. The problem with this is that most users just want to see the examples, so it feels "too prominent". Might not be necessary now we have plugin guide

en/guide/sdk.md
```diff
@@ -0,0 +1,155 @@
+# DroneCore SDKs
```
hamishwillee: @julianoes So this is supposed to be a guide on writing extension SDKs, covering just the bits that are relevant to SDK developers only.

en/guide/sdk.md
```diff
@@ -0,0 +1,155 @@
...
+
+```
+├── DroneCore
+│   ├── core
+│   ├── integration_tests
+│   └── plugins
+│       ├── action
+│       ├── ...
+│       └── telemetry
+├── SDK_Extensions
```
hamishwillee: @julianoes This is supposed to make it clear that SDK extensions are usually "peers" of dronecore.

en/guide/sdk.md
```diff
@@ -0,0 +1,155 @@
...
+  set(additional_includes "include_dir" PARENT_SCOPE)
+  ```
+* You can also add tests with `unittest_source_files`, as discussed in the following section.
+ 
+Plugins are otherwise written exactly as discussed in [Writing Plugins](../contributing/plugins.md). 
+
+
+## Testing
+
+SDK Extension tests are the same as those for the default plugins.
```
hamishwillee: @julianoes On reflection I think that tests are the same for all types of plugins - just defined in different CMakeList.txt, so this can be cut back and just point to the test docs (which may be then moved to the plugin docs)

en/guide/sdk.md
```diff
@@ -0,0 +1,155 @@
...
+To build *DroneCore* so that it includes the SDK/External plugins, specify the top level directory `EXTERNAL_DIR` in the `make` command. 
+The line below shows how this is done for the [external_example](https://github.com/dronecore/DroneCore/tree/master/external_example) directory.
+
+```
+make EXTERNAL_DIR=external_example
+```
+
+Only one external directory can be specified.
+
+<!--
```
hamishwillee: This is placeholder.

en/contributing/plugins.md
julianoes: I'm not sure I would mention the "SDK" thing.

The rest is exactly true.

en/contributing/plugins.md
julianoes: Hm. The only difference is about external additional libraries. I would say copying from an existing standard plugin makes sense.

en/contributing/test.md
```diff
@@ -1,20 +1,28 @@
 # Testing
 
-## Unit Tests
 
-To run the unit tests:
+*DroneCore* has both unit and integration tests, written using the *Google C++ Test Framework*.
```
julianoes: Maybe add `(gest)`.

en/contributing/test.md
```diff
@@ -1,20 +1,28 @@
 # Testing
 
-## Unit Tests
 
-To run the unit tests:
+*DroneCore* has both unit and integration tests, written using the *Google C++ Test Framework*. 
+The tests are run every time new code is committed to our codelines, and must pass before the code can be merged.
```
julianoes: Currently that's only true for unit tests.

en/contributing/test.md
julianoes: I would at least link to this from the plugins sections about how to add integration tests.

en/contributing/test.md
```diff
@@ -77,3 +87,138 @@ To run all telemetry tests:
...
+#include "example_impl.h"
+#include "dronecore.h"
+#include "global_include.h"
+#include <gtest/gtest.h>
+
+namespace dronecore {
+
+TEST(ExampleImpl, NoTest)
+{
+    ASSERT_TRUE(true);
```
julianoes: I guess once we have mavlink mocking, this would make more sense.

en/examples/README.md
julianoes: Ok that's good!

en/guide/sdk.md
julianoes: I would not call it SDK though. I know the Yuneec SDK is called that way but that shouldn't be the default. The goal is to have DroneCore plus extensions from a vendor (without the SDK lock in).

en/guide/sdk.md
```diff
@@ -0,0 +1,155 @@
...
+> [Testing](../contributing/test.md).
+
+
+## SDK Architecture
+
+SDK extensions are defined in a separate/parallel folder to DroneCore. 
+The extension folder *mirrors* DroneCore, and **must** contain the folders **integration_tests** and **plugins**.
+The *plugin* folder contains appropriately named sub-folders for each plugin (along with its unit test code).
+
+A simplified view of a "typical" DroneCore SDK is shown below (with extension directory: *SDK_Extensions*).
```
julianoes: I would say:

DroneCore can be extended with plugins and integration tests that are defined out of tree. This way they can be checked in to a separate repository which reduces git conflicts.

en/guide/sdk.md
julianoes: Yes, this makes sense. Just don't call it SDK.

en/guide/sdk.md
```diff
@@ -0,0 +1,155 @@
...
+For example to run the external_example hello world integration test, do:
+```
+build/default/integration_tests_runner --gtest_filter="ExternalExampleHello"
+```
+ 
+### Unit Tests
+
+Unit tests are stored in the same folder as their associated plugin source code. 
+The test file name(s) are added to the `$unittest_source_files` variable in the 
+plugin **CMakeLists.txt** file (you can add multiple files).
```
julianoes: Correct. Unit tests should not use any other dependencies from other plugins.

en/guide/sdk.md
```diff
@@ -0,0 +1,155 @@
...
+```
+make EXTERNAL_DIR=external_example
+```
+
+Only one external directory can be specified.
+
+<!-- 
+## Additional Functionality
+
+### Locking/Unlocking the SDK
```
julianoes: Fair enough, it's ok to mention it here 😄

en/contributing/test.md
hamishwillee: OK

en/contributing/test.md
hamishwillee: OK. I'll update

en/contributing/test.md
hamishwillee: Already done.

en/contributing/plugins.md
hamishwillee: I have removed. everywhere.

hamishwillee:
@julianoes 
Hi Julian

This has been re-written in line with your comments. It is now quite a lot simpler.
1. No mention of SDK. The name is "DroneCore Extensions"
2. Moved the "writing a test" into the writing a plugin doc - The testing doc points to this.
3. The DroneCore Extensions Doc is now pretty simple. The main thing it does is explain what the extensions are. 

I think this is good. Unfortunately it is going into develop - would like it to be in master, but that can't happen until the DroneCore codeline moves from develop to master, because I don't know how to cherry pick.

hamishwillee:
Thanks! Sometimes I have to write a "somewhat crap" version so I have a good place where I can test my assumptions and ask you questions. Hopefully leading to something good in the end.

PR #61, opened 2017-11-08
Add libtinyxml2-dev dependency to build instructions

hamishwillee:
As per https://github.com/dronecore/DroneCore/issues/151#issuecomment-342921001

@julianoes Anything special need to happen/be documented to add this dependency for Windows?


julianoes:
Thanks!

hamishwillee:
Thanks, BTW though, see above "@julianoes Anything special need to happen/be documented to add this dependency for Windows?"

julianoes:
No, on Windows the submodule is used and added to the static library. On Linux a dynamic library is build and the dependency tinyxml2 exists. I need to check how this pans out with the .deb file.

Issue #66, opened 2017-11-14
Update action docs for VTOL transition changes


hamishwillee:
When https://github.com/dronecore/DroneCore/pull/163 is complete do:
* [x] Update API reference
* [x] Add guide docs - probably just in the "Action" guide, but maybe worth having VTOL specific section.
* [x]  Add doc for the new example

hamishwillee:
All done.

PR #68, opened 2017-11-17
Update guide for addition of VTOL transition API

hamishwillee:
This will address #66 (captures changes in https://github.com/dronecore/DroneCore/pull/163 ) :

* [x] Update API reference
  * [x] Review output
* [x] Add guide docs - probably just in the "Action" guide, but maybe worth having VTOL specific section.
* [x]  Add doc for the new example


hamishwillee:
Hi @julianoes ,
This is now ready for review. I'll give it another check over before merging, but I think it is basically OK.  Note, also added "VTOL Support" section in guide as "call to action".

en/guide/vtol.md
```diff
@@ -0,0 +1,39 @@
...
+- Take off and land in *multicopter* mode
+- Fly in multicopter mode or transition and fly in fixed wing mode. 
+- Fly missions in the currently mode when the mission is started.
+- Use the camera/gimbal and other generic features.
+
+
+## Remaining Work
+
+Known omissions:
+- Takeoff and landing is not supported in missions.
```
julianoes: Takeoff is fine for multicopter mode. You set a mission and the takeoff waypoint is set automatically.

en/guide/vtol.md
```diff
@@ -0,0 +1,39 @@
...
+- VTOL takeoff in fixed wing mode is not supported.
+- More research is required to confirm that no other VTOL APIs are required. 
+- VTOL specific testing needs to be considered and implemented.
```
julianoes: Otherwise that's all correct, I think.

en/guide/vtol.md
```diff
@@ -0,0 +1,39 @@
...
+
+
+## Further Information
+
+Additional information/examples are linked below:
```
julianoes: Nice, I like all the helpful links.

en/examples/transition_vtol_fixed_wing.md
```diff
@@ -0,0 +1,242 @@
...
+)
+
+target_link_libraries(transition_vtol_fixed_wing
+    dronecore
+)
+```
+
+[fly_mission.cpp](https://github.com/dronecore/DroneCore/blob/{{ book.github_branch }}/example/transition_vtol_fixed_wing/transition_vtol_fixed_wing.cpp)
+
+```cpp
```
julianoes: At some point it would be nice to be able to reference and pull in source code from github, otherwise it might get outdated.

hamishwillee: Yes it would. There is a plugin that does this, but unfortunately it requires a version of node that is higher than is available on gitbook hosting. They tried to update node for me but that broke some other stuff so they had to revert. Some options:
1. I re-write the plugin to remove the ES6 promisy stuff.
2. We wait until their fancy pure-hosted solution comes out that sorts this out (but which is currently running lots behind schedule, and I am not yet sure will do everything we need)
3. I write some sort of clever auto-update script. 

I expect either 1 or 2 will happen, but not as high priority to pursue as other things on my list.

en/guide/vtol.md
hamishwillee: 1. So how does that sort of takeoff differ from [MAV_CMD_NAV_VTOL_TAKEOFF](https://mavlink.io/en/messages/common.html#MAV_CMD_NAV_VTOL_TAKEOFF)
2. After doing the MC takeoff can you send the transition command and proceed as VTOL?

en/guide/vtol.md
hamishwillee: Thanks!

en/guide/vtol.md
```diff
@@ -0,0 +1,40 @@
+# VTOL Support
+
+DroneCore has basic support for VTOL vehicles. 
+
+> **Note** Most design and test effort has gone into multicopter support. 
+> [Get in touch](../README.md#getting-help) if you would like to help enhance the VTOL/Fixed-wing experience! 
+
+## Supported Functionality
```
hamishwillee: @julianoes  Can you please re-review this section - which I updated to make it clear that MC takeoff in missions then manual transition to FW is supported, but VTOL takeoff in missions with automatic transition to FW is not.

en/examples/transition_vtol_fixed_wing.md
julianoes: Ah interesting, alright no worries.

en/guide/vtol.md
julianoes: Yep all good.

PR #73, opened 2017-12-04
Follow me guide

hamishwillee:
This adds follow me guide.

This is untested. Waiting on some of the following to be looked at:
- [ ] Follow Me has no example  - https://github.com/dronecore/DroneCore/issues/189
- [x]  Global installation not working on Ubuntu - https://github.com/dronecore/DroneCore/issues/190
- [x] #191 FollowMe API feels inconsistent - https://github.com/dronecore/DroneCore/issues/191

### DIFF
--- a/en/guide/follow_me.md
+++ b/en/guide/follow_me.md
@@ -1,0 +1,75 @@
+# Follow Me
+
+The [Follow Me](../api_reference/classdronecore_1_1_follow_me.md) class is used to engage the PX4 [Follow Me Mode](https://docs.px4.io/en/flight_modes/follow_me.html) (multicopter only). In this mode a copter will automatically yaw to face and follow a user at a specified position and distance.
+
+The API is used to supply the position(s) for the [target](../api_reference/structdronecore_1_1_follow_me_1_1_target_location.md) and the relative [follow position](../api_reference/structdronecore_1_1_follow_me_1_1_config.md) of the vehicle. Applications must get target position information from the underlying platform (or some other source). The location APIs for supported platforms are listed below:
+- Android: [Location](https://developer.android.com/reference/android/location/Location.html)
+- Apple: [Core Location Framework](https://developer.apple.com/documentation/corelocation)
+- Windows: [Windows.Devices.Geolocation](https://docs.microsoft.com/en-us/uwp/api/Windows.Devices.Geolocation)
+
+
+## Set the Follow Configuration
+
+By default the vehicle will follow directly behind the target at a height and distance of 8 metres. 
+You can (optionally) call [set_config()](../api_reference/classdronecore_1_1_follow_me.md#classdronecore_1_1_follow_me_1a4b92c3a042911dd9bdb378c686458a34) at any time to specify a different height, follow distance, relative position (front left/right/centre or behind) and responsiveness to target movements. 
+
+The code fragment below shows how to set the configuration:
+```cpp
+// configure follow me behaviour
+FollowMe::Config config;
+config.min_height_m = 12.f;  // Minimum height
+config.follow_dist_m = 20.f;  // Follow distance
+config.responsiveness = 0.2f;  // Higher responsiveness
+config.follow_direction = FollowMe::Config::FollowDirection::FRONT;  //Follow from front-centre
+
+// Apply configuration
+bool configured = device.follow_me().set_config(config);
+if (configured) {
+    // handle config failure
+    std::cout << "Configuration failed" << std::endl;
+}
+```
+
+The [get_config()](../api_reference/classdronecore_1_1_follow_me.md#classdronecore_1_1_follow_me_1a054aebafe0839a1028f277285b769fe5) method is provided to get the current configuration:
+```cpp
+auto curr_config = device.follow_me().get_config();
+```
+
+## Following a Target
+
+To start and stop following a target, call [start()](../api_reference/classdronecore_1_1_follow_me.md#classdronecore_1_1_follow_me_1a694749d43d527f85584df25a49b05ccf) and [stop()](../api_reference/classdronecore_1_1_follow_me.md#classdronecore_1_1_follow_me_1a6394507b0fb96bceebe6efd17f0529ce), respectively.
+
+In between the calls, use [set_curr_target_location()](../api_reference/classdronecore_1_1_follow_me.md#classdronecore_1_1_follow_me_1afb8c24ed93421e904b0f528569c7699a) to set the target position for the vehicle to follow. 
+DroneCore automatically resends the last set position at the rate required by the autopilot, but this can be overridden by setting the location at another rate.
+
+> **Note** The API does not automatically get the position of the device running the DroneCore. The client app must fetch this using OS-specific methods and pass them to the `FollowMe` class.
+
+```cpp
+// Start following
+FollowMe::Result follow_me_result = device.follow_me().start();
+if (follow_me_result != FollowMe::Result::SUCCESS) {
+    // handle start failure (in this case print error)
+    std::cout << "Failed to start following:" << FollowMe::result_str(follow_me_result) << std::endl;
+}
+
+// ... Get target position from underlying platform or wherever and supply to vehicle 
+// Here we just show one point being set. Minimum values specified in the configuration are respected.
+follow_me.set_curr_target_location({ 47.39776569, 8.54553292, 9.0, 0.f, 0.f, 0.f });
+
+// Stop following
+follow_me_result = device.follow_me().stop();
+if (follow_me_result != FollowMe::Result::SUCCESS) {
+    // handle stop failure (in this case print error)
+    std::cout << "Failed to stop following:" << FollowMe::result_str(follow_me_result) << std::endl;
+}
+```
+
+The last location that was set can be retrieved using [get_last_location()](../api_reference/classdronecore_1_1_follow_me.md#classdronecore_1_1_follow_me_1ab68273d5ace65ee953afa1797ae49e7c).
+
+
+## Further Information
+
+Additional information/examples for the Follow Me API are linked below:
+
+* Integration tests:
+  * [follow_me.cpp](https://github.com/dronecore/DroneCore/blob/{{ book.github_branch }}/integration_tests/follow_me.cpp)
--- a/en/guide/follow_me.md
+++ b/en/guide/follow_me.md
@@ -52,9 +52,9 @@ if (follow_me_result != FollowMe::Result::SUCCESS) {
 }
 
 
-// ... Get target position from underlying platform or wherever and supply to vehicle 
-// Here we just show one point being set. Minimum values specified in the configuration are respected.
-follow_me.set_target_location({ 47.39776569, 8.54553292, 9.0, 0.f, 0.f, 0.f });
+// Get target position from underlying platform and supply to vehicle. 
+//   For this example we just show one point being set (instead of a stream).
+follow_me.set_target_location({ 47.39776569, 8.54553292, 0.f, 0.f, 0.f, 0.f });
 
 
 // Stop following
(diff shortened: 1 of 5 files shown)


en/guide/follow_me.md
```diff
@@ -0,0 +1,75 @@
...
+
+// Stop following
+follow_me_result = device.follow_me().stop();
+if (follow_me_result != FollowMe::Result::SUCCESS) {
+    // handle stop failure (in this case print error)
+    std::cout << "Failed to stop following:" << FollowMe::result_str(follow_me_result) << std::endl;
+}
+```
+
+The last location that was set can be retrieved using [get_last_location()](../api_reference/classdronecore_1_1_follow_me.md#classdronecore_1_1_follow_me_1ab68273d5ace65ee953afa1797ae49e7c).
```
hamishwillee: @shakthi-prashanth-m What happens if you haven't set any location when you select `start()`? Do you have to set a start location before starting the API.

en/guide/follow_me.md
```diff
@@ -0,0 +1,75 @@
...
+```cpp
+auto curr_config = device.follow_me().get_config();
+```
+
+## Following a Target
+
+To start and stop following a target, call [start()](../api_reference/classdronecore_1_1_follow_me.md#classdronecore_1_1_follow_me_1a694749d43d527f85584df25a49b05ccf) and [stop()](../api_reference/classdronecore_1_1_follow_me.md#classdronecore_1_1_follow_me_1a6394507b0fb96bceebe6efd17f0529ce), respectively.
+
+In between the calls, use [set_curr_target_location()](../api_reference/classdronecore_1_1_follow_me.md#classdronecore_1_1_follow_me_1afb8c24ed93421e904b0f528569c7699a) to set the target position for the vehicle to follow. 
+DroneCore automatically resends the last set position at the rate required by the autopilot, but this can be overridden by setting the location at another rate.
```
hamishwillee: @shakthi-prashanth-m Is this correct? I assume so, but want to make sure!

en/guide/follow_me.md
```diff
@@ -0,0 +1,75 @@
...
+    std::cout << "Failed to start following:" << FollowMe::result_str(follow_me_result) << std::endl;
+}
+
+// ... Get target position from underlying platform or wherever and supply to vehicle 
+// Here we just show one point being set. Minimum values specified in the configuration are respected.
+follow_me.set_curr_target_location({ 47.39776569, 8.54553292, 9.0, 0.f, 0.f, 0.f });
+
+// Stop following
+follow_me_result = device.follow_me().stop();
+if (follow_me_result != FollowMe::Result::SUCCESS) {
```
hamishwillee: @shakthi-prashanth-m @julianoes Can this ever happen? ie if say we lost connection, won't the vehicle stop getting updates from DroneCore, so you'll stop following anyway?

What happens when you stop following? Does the mode change? What will the vehicle do - go to hold mode?

en/guide/follow_me.md
shakthi-prashanth-m: @hamishwillee It is optional to set location before starting FollowMe; in that case, `get_last_location()` would return `NAN` values. In the [second integration test of FollowMe](https://github.com/dronecore/DroneCore/blob/develop/integration_tests/follow_me.cpp#L79-L147), there's no location set at all.

en/guide/follow_me.md
shakthi-prashanth-m: Yes. but rate remains same. I meant this **DroneCore keeps sending last set position at rate 1Hz until client app overrides by setting new location.**

en/guide/follow_me.md
shakthi-prashanth-m: @hamishwillee thats a good point. Thanks.
I simulated this scenario by pressing Ctrl-C in FollowMe mode. I monitored the mode for long time like after 2 mins or so, it remains in `AUTO.FOLLOW_TARGET`. I did see a [MAVLink command](http://mavlink.org/messages/common/#MAV_CMD_DO_FOLLOW) that controls how long a vehicle should be in FollowMe or FOLLOW_TARGET mode. But this command execution failed when I executed using MAVROS command-line.  I feel this should be part of FollowMe configuration; a PX4 parameter can be used to configure this value similar to other FollowMe configuration parameters. It is important to know how to use the above MAVLink command.

en/guide/follow_me.md
hamishwillee: Thanks. So what actually happens - the API doesn't send anything, or PX4 just ignores whatever messages it gets?

en/guide/follow_me.md
hamishwillee: Thanks. I'll fix that text.

en/guide/follow_me.md
hamishwillee: @shakthi-prashanth-m . Hmmm, so the "intention" is that the vehicle go to hold mode when stop is called or after some timeout from having lost connection. I think you're probably right to provide this as a configuration value with a sensible default value (probably as part of existing `Config`) - because I suspect that otherwise this will end up being a semi-random value. What do you think @julianoes 

I'd be setting the parameter rather than this mavlink message. Did you try that?

en/guide/follow_me.md
shakthi-prashanth-m: When we start FollowMe => We send `MAV_CMD_DO_SET_MODE` to change to `PX4_CUSTOM_SUB_MODE_AUTO_FOLLOW_TARGET` mode.
If there's no position set, nothing happens at Vehicle side other than being at FollowMe mode.
When we start pushing GPS location updates, they're sent to Vehicle.

en/guide/follow_me.md
shakthi-prashanth-m: @hamishwillee , I would like to elaborate my explanation further:

When `FollowMe::stop()` is called, [we change](https://github.com/dronecore/DroneCore/blob/develop/plugins/follow_me/follow_me_impl.cpp#L167) the Vehicle mode to LOITER.
However, if we lost connection to Vehicle abruptly as you rightly pointed out, then the Vehicle remains at the previously published location in `AUTO.FOLLOW_TARGET` mode.
Like I said earlier, there is a MAVLink command to configure TTL seconds (duration in seconds for which Vehicle stays in FollowMe mode), and then switches to default hold mode. But I tried to use this command, it doesn't seem to work. I need to figure out how to use this command. Please let me know if you know it.

As I see there's no PX4 parameter to configure this TTL seconds value.
I didn't get your point about setting parameter, as there's none.

hamishwillee: @shakthi-prashanth-m Thanks for clarification.

I don't know if I'm looking in the right place but the [navigator code here](https://github.com/PX4/Firmware/blob/master/src/modules/navigator/follow_target.cpp#L344) seems to indicate that the vehicle will just wait forever (well, some failsafe) if no setpoint is received. In other words, I don't think the timeout has been implemented. I think we should raise a bug for that in PX4, and ideally fix it.

The fix should include a parameter that can be used to set the timeout - whether or not DroneCore actually calls an API to change that parameter is a different discussion.

Does that make sense?

en/guide/follow_me.md
hamishwillee: Thanks!

hamishwillee:
FYI, just updated this to clarify that when stop is called we go into Hold mode, and what happens if no position is set.

en/guide/follow_me.md
```diff
@@ -52,10 +51,12 @@ if (follow_me_result != FollowMe::Result::SUCCESS) {
     std::cout << "Failed to start following:" << FollowMe::result_str(follow_me_result) << std::endl;
 }
 
+
 // ... Get target position from underlying platform or wherever and supply to vehicle 
 // Here we just show one point being set. Minimum values specified in the configuration are respected.
 follow_me.set_curr_target_location({ 47.39776569, 8.54553292, 9.0, 0.f, 0.f, 0.f });
```
shakthi-prashanth-m: @hamishwillee altitude here is not of significance. This is altitude of the target. We've already specified Vehicle's altitude in FollowMe configuration.

en/guide/follow_me.md
shakthi-prashanth-m: @hamishwillee  you're welcome.
I too vote for having PX4 parameter say `NAV_FT_TTL` (in secs) which decides how long the vehicle is in Follow Target mode (waiting for messages).

en/guide/follow_me.md
hamishwillee: Yes, thanks.

en/guide/follow_me.md
hamishwillee: Yes. Though I'd probably suggest we don't actually set it in the API (just document it).

hamishwillee: I assume you are pursuing that :-)

hamishwillee:
@julianoes @shakthi-prashanth-m This is up to date with the current API ref. Would be nice to have an example, but I think is probably good to merge now "as is". Can you please review?

en/guide/follow_me.md
```diff
@@ -0,0 +1,77 @@
...
+FollowMe::Result follow_me_result = device.follow_me().start();
+if (follow_me_result != FollowMe::Result::SUCCESS) {
+    // handle start failure (in this case print error)
+    std::cout << "Failed to start following:" << FollowMe::result_str(follow_me_result) << std::endl;
+}
+
+
+// ... Get target position from underlying platform or wherever and supply to vehicle 
+// Here we just show one point being set. Minimum values specified in the configuration are respected.
+follow_me.set_target_location({ 47.39776569, 8.54553292, 9.0, 0.f, 0.f, 0.f });
```
shakthi-prashanth-m: It should be `follow_me.set_target_location({ 47.39776569, 8.54553292, 0., 0.f, 0.f, 0.f });`

en/guide/follow_me.md
```diff
@@ -0,0 +1,77 @@
...
+The last location that was set can be retrieved using [get_last_location()](../api_reference/classdronecore_1_1_follow_me.md#classdronecore_1_1_follow_me_1a16da2bf7d0384e2bff4440600b523f8c). Before a target position is first set this API will return `Nan`.
+
+
+
+## Further Information
+
+Additional information/examples for the Follow Me API are linked below:
+
+* Integration tests:
+  * [follow_me.cpp](https://github.com/dronecore/DroneCore/blob/{{ book.github_branch }}/integration_tests/follow_me.cpp)
```
shakthi-prashanth-m: This link didn't work.
Correct one: https://github.com/dronecore/DroneCore/blob/develop/integration_tests/follow_me.cpp

shakthi-prashanth-m:
Links to FollowMe::Result, FollowMe::Config::FollowDirection, etc were not opening. Can this be ignored, as it should be working in the final doc ..

en/guide/follow_me.md
hamishwillee: This is not an issue. The `{{ book.github_branch }}` text dynamically inserts `develop` or `master` based on a variable set in the **book.json**. This is a way for me to make sure that our links are correct for a particular version.

en/guide/follow_me.md
hamishwillee: Thanks. I will change to this:
```
// Get target position from underlying platform and supply to vehicle. 
//   Here we just show one point being set.
follow_me.set_target_location({ 47.39776569, 8.54553292, 0.f, 0.f, 0.f, 0.f });
```
I realised that saying  "Minimum values specified in the configuration are respected." is irrelevant here - since we're not specifying the follow height, but the target height. Configuration settings are irrelevant.

hamishwillee:
I'm going to fix up and merge this. I do have one associated query related to config though

hamishwillee:
> Links to FollowMe::Result, FollowMe::Config::FollowDirection, etc were not opening. Can this be ignored, as it should be working in the final doc ..

Yes, it can be ignored. 

FYI, The problem is that github does not generate proper heading anchors based on the `{#anchor_name}` syntax after the heading. What it is supposed to do is use that string. What it actually does is append that string to the existing heading anchor. A bug IMO.

en/guide/follow_me.md
shakthi-prashanth-m: @hamishwillee created [issue in PX4](https://github.com/PX4/Firmware/issues/8501) for adding this parameter.

PR #74, opened 2017-12-04
Add enable/disable to plugin guide

hamishwillee:


en/contributing/plugins.md
```diff
@@ -68,6 +68,54 @@ The [standard plugins](https://github.com/dronecore/DroneCore/tree/{{ book.githu
...
+* [init()](#init)/[deinit()](#deinit): These are called when a device is created and just before it is destroyed. These should be used for setting up and cleaning everything that depends on having the `Device` instantiated. This includes calls that set up callbacks.
+* [enable()](#enable)/[disable()](#disable): These are called when a vehicle is discovered or has timed out. They should be used for managing resources needed to access a connected device/vehicle (e.g. getting a parameter or changing a setting).
+
+
+> **Note** The methods are currently optional. For a transitional period, runtime warnings will be displayed if the methods are not implemented.
+
+Additional detail is provided for methods below.
+
+
+##### init() {#init}
```
hamishwillee: I pulled in your docs here because the PluginImplBase  is not part of the public API, and hence not automatically exported and build into the API Reference. 
Should it be? How can we make that happen?

IF that happens I'd remove these docs here and link to the ref. That would be better.

en/contributing/plugins.md
```diff
@@ -68,6 +68,54 @@ The [standard plugins](https://github.com/dronecore/DroneCore/tree/{{ book.githu
...
+### Plugin Enable/Disable
+
+DroneCore provides virtual methods that a plugin should implement allow DroneCore to better manage resources. For example, to prevent callback being created before the `Device` is instantiated, or messages being sent when a vehicle is not connected.
+
+Plugin authors should provide an implementation class that derives from `PluginImplBase` (**core/plugin_impl_base.h**) and implements the following pure virtual methods:
+* [init()](#init)/[deinit()](#deinit): These are called when a device is created and just before it is destroyed. These should be used for setting up and cleaning everything that depends on having the `Device` instantiated. This includes calls that set up callbacks.
+* [enable()](#enable)/[disable()](#disable): These are called when a vehicle is discovered or has timed out. They should be used for managing resources needed to access a connected device/vehicle (e.g. getting a parameter or changing a setting).
+
+
+> **Note** The methods are currently optional. For a transitional period, runtime warnings will be displayed if the methods are not implemented.
```
hamishwillee: We say these are optional, but the methods are pure-virtuals - doesn't that mean that everyone will have to implement them in their plugins (since we expect all plugins to derive from the base?)

On that note, if we do expect everyone to derive from PluginImplBase that should be a bullet point in the section "Create a Plugin". THoughts?

hamishwillee:
I am not sure where to record what the user sees. I think the https://docs.dronecore.io/v/develop/en/guide/general_usage.html#error-handling section is sufficient.

en/contributing/plugins.md
julianoes: This will change in the future as dronecore becomes it's own library that libraries link to. At that point there will be a public API for it to be used by plugin developers.

en/contributing/plugins.md
julianoes: Oh, this comment is outdated. They are no longer optional.

en/contributing/plugins.md
hamishwillee: OK. I'll update that as a PR when those changes go in.

en/contributing/plugins.md
hamishwillee: Will remove.

en/contributing/plugins.md
```diff
@@ -68,6 +68,58 @@ The [standard plugins](https://github.com/dronecore/DroneCore/tree/{{ book.githu
 how to write plugin code, including how to send and process MAVLink messages.
 
 
+### Plugin Base Class
```
hamishwillee: @julianoes This is new section. I added to provide a place for talking more about the code in future, if we so wish.

en/contributing/plugins.md
```diff
@@ -68,6 +68,58 @@ The [standard plugins](https://github.com/dronecore/DroneCore/tree/{{ book.githu
...
+
+### Plugin Enable/Disable
+
+DroneCore provides virtual methods that a plugin should implement allow DroneCore to better manage resources. For example, to prevent callback being created before the `Device` is instantiated, or messages being sent when a vehicle is not connected.
+
+Plugin authors should provide an implementation of the following `PluginImplBase` pure virtual methods:
+* [init()](#init)/[deinit()](#deinit): These are called when a device is created and just before it is destroyed. These should be used for setting up and cleaning everything that depends on having the `Device` instantiated. This includes calls that set up callbacks.
+* [enable()](#enable)/[disable()](#disable): These are called when a vehicle is discovered or has timed out. They should be used for managing resources needed to access a connected device/vehicle (e.g. getting a parameter or changing a setting).
+
+The [external example](https://github.com/dronecore/DroneCore/tree/{{ book.github_branch }}/external_example) provides a minimal implementation.
```
hamishwillee: This line is new.

hamishwillee:
@julianoes OK, I have updated this to remove the note that the enable/disable methods are optional. I also made other minor changes. Can you review. I'll update to remove the "inline API reference" once the rearchitecting of the code is complete and the plugin base class is building API reference.

en/contributing/plugins.md
julianoes: That makes sense. There is not much now but there probably will be more in the future.

PR #86, opened 2018-02-01
Update build.md

hamishwillee:
Make it clear how to select master vs develop branch, and that develop is used by default.


julianoes:
Ok that's better, thanks @hamishwillee .

hamishwillee:
@julianoes A bit better but not as concise. I'm looking forward to when we can point people to binary builds as the "normal case" :-)

julianoes:
Right!

PR #88, opened 2018-02-05
Update docs to new plugin archicture

hamishwillee:
This PR updates docs to match up with https://github.com/dronecore/DroneCore/pull/188

Examples and most of guide done (though there will be further comments inline). This also includes docs from https://github.com/dronecore/DroneCore/pull/242 (not yet merged). Writing plugins and extensions not yet done. Some outstanding questions/actions

1. DevicePluginContainer removed. Need to check no longer referenced.
2. Confirm whether PluginBase, and MavlinkHandlerTableEntry part of the public API? If so need docs for those.
3. Device needs docs: https://github.com/dronecore/DroneCore/issues/243

@julianoes Even though this is not "complete", provided it is accurate as is, it might be worth accepting and then doing how to write plugins etc as separate PRs.


en/getting_started/faq.md
```diff
@@ -24,7 +24,7 @@ The architecture has stubs for serial, TCP, and UDP connections. However, only U
 
 ### Why is libCURL a dependency?
 
-libCURL will be required to download the camera definition file which is referenced in [CAMERA_INFORMATION](http://mavlink.org/messages/common#CAMERA_INFORMATION). It might also come in handy whenever any other REST resources are required.
+libCURL will be required to download the camera definition file which is referenced in [CAMERA_INFORMATION](http://mavlink.org/messages/common#CAMERA_INFORMATION). It might also come in handy whenever any other REST resources are required.
```
hamishwillee: @julianoes  Where does DroneCore use CAMERA_INFORMATION?

en/README.md
```diff
@@ -64,7 +64,6 @@ The most important classes are:
 - [Gimbal](/api_reference/classdronecore_1_1_gimbal.md): Control a gimbal.
 - [FollowMe](/api_reference/classdronecore_1_1_follow_me.md): Drone tracks a position supplied by DroneCore.
 - [Logging](/api_reference/classdronecore_1_1_logging.md): Data logging and streaming from the vehicle.
-- [include/device_plugin_container.h.in](https://github.com/dronecore/DroneCore/blob/{{ book.github_branch }}/include/device_plugin_container.h.in): Auto-generated file that is required for DroneCore plugin development - see [DevicePluginContainer](/api_reference/classdronecore_1_1_device_plugin_container.md).
```
hamishwillee: Deleted the DevicePluginContainer. Need to know if need to add: PluginBase, and MavlinkHandlerTableEntry

en/SUMMARY.md
```diff
@@ -6,9 +6,10 @@
 * [Guide](guide/README.md)
   * [Usage/Paradigms](guide/general_usage.md)
   * [Connecting to Devices](guide/connections.md)
+  * [Managing Devices](guide/using_plugins.md)
   * [Vehicle Information](guide/device_version.md)
   * [Telemetry](guide/telemetry.md)
-  * [Taking Off and Landing](guide/taking_off_landing.md)
+  * [Actions - Take Off, Arm, ...](guide/taking_off_landing.md)
```
hamishwillee: Renamed to focus on "Actions" old name too specific for broad content of topic.

en/api_reference/classdronecore_1_1_plugin_base.md
```diff
@@ -0,0 +1,58 @@
+# dronecore::PluginBase Class Reference
```
hamishwillee: @julianoes If this is public, need some docs.

en/guide/device_version.md
```diff
@@ -6,24 +6,35 @@ The [Info](../api_reference/classdronecore_1_1_info.md) class is used to get dev
 hardware may return garbage values (for example, the simulator provides garbage values for the vendor 
 firmware semantic version).
 
-The code below shows how to query the UUID, version, and product, information and print it to the console:
+## Preconditions
 
+The following code assumes that you already have included DroneCore (`#include <dronecore/dronecore.h>`) and that there is a [connection to a device](../guide/connections.md) obtained as shown below:
```
hamishwillee: @julianoes Conscious decision here NOT to include mention of also having to link to the plugin library. Do you think this has to be mentioned? We do link to the using plugins doc that covers this sort of thing.

hamishwillee:
@julianoes What needs to change for local installation?
 https://docs.dronecore.io/v/develop/en/guide/toolchain.html#dronecore_local_install

Do you think you could provide an example? I suspect it is just addition of  the plugin libs to `${dronecore_lib}`

en/README.md
julianoes: Not just yet, I think.

en/SUMMARY.md
julianoes: Makes sense.

en/examples/offboard_velocity.md
```diff
@@ -152,16 +161,18 @@ inline void offboard_log(const std::string &offb_mode, const std::string msg)
  */
 bool offb_ctrl_ned(Device &device)
 {
+    std::shared_ptr<Offboard> offboard = std::make_shared<Offboard>(&device);
```
julianoes: This is actually wrong. We should not create `offboard` twice but only once and then pass it on as `std::shared_ptr<Offboard> offboard` instead of `Device &device`.

I guess we need to fix the example first.

en/getting_started/faq.md
julianoes: Hehe, that's still proprietary but I'll open source that camera plugin soon.

en/guide/device_version.md
julianoes: You now need to include and link to each plugin that you require. So in this case we do require info.

en/examples/offboard_velocity.md
hamishwillee: Update to example is here: https://github.com/dronecore/DroneCore/pull/246
Untested because I can't build dronecore https://github.com/dronecore/DroneCore/issues/245

I'll update docs shortly.

en/getting_started/faq.md
hamishwillee: Cool. Created https://github.com/dronecore/docs/issues/90 just to make sure this doesn't fall off the radar

en/guide/device_version.md
hamishwillee: I had hoped that having that in the "Using Plugins" section would be sufficient (since all the sections otherwise have to duplicate similar instructions). Anyway, I've updated so that this is in every docs as "Create the Plugin" (rather than preconditions). OK?

hamishwillee:
OK to merge this @julianoes ? There is more to be done - like updating for the Offboard example (https://github.com/dronecore/DroneCore/pull/246), and looking at the external example, but I'd like to get this in.

hamishwillee:
Thanks!

PR #89, opened 2018-02-06
Make clear statement about indoor use support

hamishwillee:
Addresses https://github.com/dronecore/DroneCore/issues/234#issuecomment-363008920


hamishwillee:
Am merging this - the merge error is incorrect.

Issue #85, opened 2018-01-31
Building Source · Need to retest docker commands


hamishwillee:
[Building Source > Building in Docker](https://docs.dronecore.io/v/develop/en/contributing/build.html#building-in-docker) was written when we just had one Dockerfile in root. Now there are multiple files here:  https://github.com/dronecore/DroneCore/tree/develop/docker

Need to confirm instructions still OK.

julianoes:
Nice catch, there are now two dockerfiles: https://github.com/dronecore/DroneCore/tree/master/docker

One for Ubuntu and one for Fedora. To build you know need to specify which file:

```
docker build -f docker/Dockerfile-Fedora-27 -t my_image .
```

hamishwillee:
Done!

PR #91, opened 2018-02-07
Updates for new docker script info

hamishwillee:
We now have multiple docker files, so instructions for building in docker need to change - as per https://github.com/dronecore/docs/issues/85#issuecomment-362283401

@julianoes This has some issues/things that don't work. I need help! Will post inline

### DIFF
--- a/en/contributing/build.md
+++ b/en/contributing/build.md
@@ -209,7 +209,7 @@ make ios_simulator install
 
 ## Building in Docker
 
-You can also build DroneCore on your host computer with a toolchain running in a [Docker](https://docs.docker.com/) container (this saves you from having to install or manage the toolchain directly).
+You can also build DroneCore on your host computer with a toolchain running in a [Docker](https://docs.docker.com/) container (this saves you from having to install or manage the toolchain directly). There are containers based on Fedora and Ubuntu (it doesn't matter which you use).
 
 The main steps are:
 
@@ -220,17 +220,22 @@ The main steps are:
    cd DroneCore
    git submodule update --init --recursive
    ```
-1. Enter the following command in your host's terminal:
-   ```sh
-   docker run --rm -it -v $HOME/<path-to-dronecore-repo>/DroneCore:/home/docker1000/src/DroneCore:rw dronecore/dronecore bash
-   ```
+1. Enter either of the following command in your host's terminal:
+   * Fedora
+     ```sh
+     docker run --rm -it -v $HOME/DroneCore:/home/docker1000/src/DroneCore:rw dronecore/dronecore-fedora-27 bash
+     ```
+   * Ubuntu
+     ```sh
+     docker run --rm -it -v $HOME/DroneCore:/home/docker1000/src/DroneCore:rw dronecore/dronecore-ubuntu-16.04 bash
+     ```
    > **Note** The `-v` flag maps a directory on your host (left side) to a path in
    the container (right side). Above you need to specify the left-side path to the DroneCore repository on your host. The container path must be set as above.
 
-   > **Note** The `-v` flag maps a directory on your host (left side) to a path in the container (right side).  You need to specify the left-side path to the DroneCore repository on your host and the container path must be set as above.
+   > **Note** The `-v` flag maps a directory on your host (left side) to a path in the container (right side).  You need to specify the left-side path to the DroneCore repository on your host and the container path must be set as above. 
    > The `--rm` automatically cleans up leftover docker containers after you exit the docker container.
 
-   Docker will download an image from [Docker Hub](https://hub.docker.com/r/dronecore/dronecore/), use it to create a container, and then open a bash prompt:
+   Docker will download an image from [Docker Hub](https://hub.docker.com/u/dronecore/), use it to create a container, and then open a bash prompt:
    ```
    root*81ebe14d0c1a:/home/docker1000/src/Dronecore#
    ```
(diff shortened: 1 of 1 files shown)


en/contributing/build.md
```diff
@@ -220,17 +220,22 @@ The main steps are:
    cd DroneCore
    git submodule update --init --recursive
    ```
-1. Enter the following command in your host's terminal:
-   ```sh
-   docker run --rm -it -v $HOME/<path-to-dronecore-repo>/DroneCore:/home/docker1000/src/DroneCore:rw dronecore/dronecore bash
-   ```
+1. Enter either of the following command in your host's terminal:
```
hamishwillee: These just replace the old target `dronecore/dronecore` with `dronecore/dronecore-fedora-27`. The problem is that they don't work - new bash prompt is empty/not mapped to the host tree. 

I tried the following:
```
docker run --rm -it -v $HOME/DroneCore:/root/DroneCore:rw dronecore/dronecore-ubuntu-16.04 bash
```
That worked (ie the directory was mapped properly) but ownership of files is by `root`, so no access in the parent. 

What is the right way to run these two?

en/contributing/build.md
```diff
@@ -220,17 +220,22 @@ The main steps are:
...
    the container (right side). Above you need to specify the left-side path to the DroneCore repository on your host. The container path must be set as above.
 
-   > **Note** The `-v` flag maps a directory on your host (left side) to a path in the container (right side).  You need to specify the left-side path to the DroneCore repository on your host and the container path must be set as above.
+   > **Note** The `-v` flag maps a directory on your host (left side) to a path in the container (right side).  You need to specify the left-side path to the DroneCore repository on your host and the container path must be set as above. 
    > The `--rm` automatically cleans up leftover docker containers after you exit the docker container.
 
-   Docker will download an image from [Docker Hub](https://hub.docker.com/r/dronecore/dronecore/), use it to create a container, and then open a bash prompt:
+   Docker will download an image from [Docker Hub](https://hub.docker.com/u/dronecore/), use it to create a container, and then open a bash prompt:
    ```
    root*81ebe14d0c1a:/home/docker1000/src/Dronecore#
```
hamishwillee: So yes, I don't get this mapping - instead I get something like: `[root@06e7da84c095 DroneCore]#`

en/contributing/build.md
```diff
@@ -264,16 +269,22 @@ docker run --rm -it -v $HOME/<path-to-dronecore-repo>/DroneCore:/home/docker1000
...
 1. Open a command prompt/terminal in the root of the DroneCore repository.
-1. Build the image as shown:
-   ```sh
-   docker build . -t my_image    # 'my_image' can then be used to refer to the image
-   ```
-1. Open a bash prompt using this image:
+1. Build the images as shown:
+   * Fedora 
+     ```sh
+     docker build -f docker/Dockerfile-Fedora-27 -t my_image .
```
hamishwillee: In my Ubuntu VM this fails with this gist: https://gist.github.com/hamishwillee/906f18ed8c8a81231d8c4efb5afa7c81

en/contributing/build.md
```diff
@@ -264,16 +269,22 @@ docker run --rm -it -v $HOME/<path-to-dronecore-repo>/DroneCore:/home/docker1000
...
+     ```
+   * Ubuntu 
+     ```sh
+     docker build -f docker/Dockerfile-Ubuntu-16.04 -t my_image .
```
hamishwillee: In my Ubuntu VM this Ubuntu dockerimage build fails with this gist: https://gist.github.com/hamishwillee/a41e2a76502b63f2926d3c6a698b13a9

en/contributing/build.md
julianoes: Yes, the ownership problem is a general issue with docker and a bit tricky. I usually just do a `sudo make clean` outside of docker afterwards, or I remember to do `make clean` before leaving docker.

en/contributing/build.md
hamishwillee: @julianoes Any idea why both of these fail? I can't publish a section on how to build the image it doesn't work.

en/contributing/build.md
hamishwillee: PS NOte that this may also be why the files are mapped to -v `$HOME/DroneCore:/home/docker1000/src/DroneCore:rw` - because root/Dronecore belongs to root.

hamishwillee:
@julianoes I've updated this so it can be used to build using the docker images. While I'd love for us to [support passing in user id](https://github.com/dronecore/docs/pull/91/commits/0b2e72f86122b5b9f94e275cbc42348e4e011dec#r166797064) that is not a blocker (should I create issue requesting that change?)

What is blocking this is that building the images fails. Does this work for you?

en/contributing/build.md
julianoes: Oh, that's right, it is now `make distclean` to remove the actual folders.

julianoes: > Could we update the scripts to support passing in the current user? Px4 docker allows you to pass in the current user ID as --env=LOCAL_USER_ID="$(id -u)" - this is then used to create the files. You can see this being loaded into the environment here

I would like to do it a bit cleaner than that but I need to look into it.
I made an issue for now: #254

en/contributing/build.md
julianoes: Ok, that's kind of expected. With upstream gRPC changes. Since we have our own gRPC fork anyway, we can remove this kind of stuff.

hamishwillee: Great, so this should be buildable "sometime soon"?

julianoes: See https://github.com/dronecore/DroneCore/pull/255

hamishwillee:
@julianoes Any reason not to merge this? 

The instructions are all correct so even if some thing don't work this should be published (and actually building the docker images now works thanks). I'll update again when/if if we add support for passing in the user id to docker.

hamishwillee:
@julianoes Thanks. FYI The CI fails are non-blocking external link test fails. I'll disable them as a fail :-)

PR #98, opened 2018-02-15
build: added how to use all CPU cores to build

julianoes:
On Windows, you apparently need to set this env variable to speed up the
build.


en/contributing/build.md
```diff
@@ -111,6 +111,11 @@ To build the *DroneCore* C++ Library on Windows:
      ```sh
      cmake --build . --config Release
      ```
+   <span></span>
```
hamishwillee: I added this span. Due to bugs in gitbook it is needed if you render two notes/tips/warnings immediately after each other.

Issue #99, opened 2018-02-15
Add new docs updates for QGC Mission import


hamishwillee:
Merged here: https://github.com/dronecore/DroneCore/pull/235

hamishwillee:
Image of mission here: https://github.com/dronecore/DroneCore/pull/235#issuecomment-365863674
Fly mission console output here: https://github.com/dronecore/DroneCore/pull/235#issuecomment-365867957

hamishwillee:
This will be fixed by https://github.com/dronecore/docs/pull/100

PR #100, opened 2018-02-16
[WIP] Update docs following addition of QGC plan import

hamishwillee:
Updates docs for new QGC plan file importer (https://github.com/dronecore/DroneCore/pull/235) as requested in #99.

Actions
* [x] API Ref changes
* [x] New example
* [x] Update guide with new APIs. Cross link to new tests.


hamishwillee:
This is still WIP because:
* [ ] Waiting on https://github.com/dronecore/DroneCore/pull/271 to be approved (may result in API ref updates)
* [ ] Missing sample plan file: https://github.com/dronecore/DroneCore/issues/272 - may result in changes to the example - e.g. at the moment there is a placeholder in the example page, but may need to remove that.

en/examples/fly_mission_qgc_plan.md
```diff
@@ -0,0 +1,337 @@
...
+                  << NORMAL_CONSOLE_TEXT << std::endl;
+        exit(EXIT_FAILURE);
+    }
+}
+```
+
+[qgroundcontrol_sample.plan](https://github.com/dronecore/DroneCore/blob/{{ book.github_branch }}/example/fly_qgc_mission/qgroundcontrol_sample.plan)
+
+```
+
```
hamishwillee: This is empty while I wait to see where the example plan file lives - https://github.com/dronecore/DroneCore/issues/272

If it stays in the plugin then I update the cpp file and delete this block.

shakthi-prashanth-m:
@hamishwillee, do you need full screenshot of the QGC plan for higher resolution ? I did crop the image just enough to cover mission items while uploading.

en/guide/missions.md
```diff
@@ -1,6 +1,6 @@
 # Missions
 
-The DroneCore Mission API allows you to create, upload, download, run, pause, restart, jump to item in, and track missions. Missions can have multiple "mission items", each which may specify a position, altitude, fly-through behaviour, camera action, gimbal position, and the speed to use when traveling to the next position.
+The DroneCore Mission API allows you to create, upload, download, import from *QGroundControl*, run, pause, restart, jump to item in, and track missions. Missions can have multiple "mission items", each which may specify a position, altitude, fly-through behaviour, camera action, gimbal position, and the speed to use when traveling to the next position.
```
shakthi-prashanth-m: >Missions can have multiple "mission items"

Shouldn't this be **A mission can have multiple "mission items"** ?

en/guide/missions.md
```diff
@@ -10,13 +10,14 @@ Missions are *managed* though the [Mission](../api_reference/classdronecore_1_1_
...
-* [MAV_CMD_VIDEO_START_CAPTURE](http://mavlink.org/messages/common#MAV_CMD_VIDEO_START_CAPTURE)
-* [MAV_CMD_VIDEO_STOP_CAPTURE](http://mavlink.org/messages/common#MAV_CMD_VIDEO_STOP_CAPTURE)
+* [MAV_CMD_NAV_WAYPOINT](https://mavlink.io/en/messages/common.html#MAV_CMD_NAV_WAYPOINT)
+* [MAV_CMD_DO_CHANGE_SPEED](https://mavlink.io/en/messages/common.html#MAV_CMD_DO_CHANGE_SPEED)
+* [MAV_CMD_DO_MOUNT_CONTROL](https://mavlink.io/en/messages/common.html#MAV_CMD_DO_MOUNT_CONTROL)
+* [MAV_CMD_IMAGE_START_CAPTURE](https://mavlink.io/en/messages/common.html#MAV_CMD_IMAGE_START_CAPTURE)
+* [MAV_CMD_IMAGE_STOP_CAPTURE](https://mavlink.io/en/messages/common.html#MAV_CMD_IMAGE_STOP_CAPTURE)
+* [MAV_CMD_VIDEO_START_CAPTURE](https://mavlink.io/en/messages/common.html#MAV_CMD_VIDEO_START_CAPTURE)
+* [MAV_CMD_VIDEO_STOP_CAPTURE](https://mavlink.io/en/messages/common.html#MAV_CMD_VIDEO_STOP_CAPTURE)
+* [MAV_CMD_NAV_LOITER_TIME](https://mavlink.io/en/messages/common.html#MAV_CMD_NAV_LOITER_TIME)
```
shakthi-prashanth-m: Don't we need to mention first mission item in QGroundControl plan:  `MAV_CMD_NAV_TAKEOFF` ? OR Is it implicit as part of `MAV_CMD_NAV_WAYPOINT` ?

en/guide/missions.md
```diff
@@ -126,18 +128,37 @@ std::shared_ptr<MissionItem> add_mission_item(double latitude_deg,
...
+                      MissionItem::CameraAction::NONE));
 ```
 
+## Import a Mission from a QGC Plan {#import_qgc_plan}
+
+DroneCore allows you to import a mission from a *QGroundControl* plan (the imported mission can then be uploaded to a vehicle).
+
+> **Note** To export a mission plan from the *QGroundControl* use the [Sync Tool](https://docs.qgroundcontrol.com/en/PlanView/PlanView.html#sync) (**Plan View > Sync Tool**, and then select **Save to File**).
+
+The mission is imported using the static [import_qgroundcontrol_mission](../api_reference/classdronecore_1_1_mission.md#classdronecore_1_1_mission_1a7c73e97e5c1395a7451bb659d03e5f57) method. The method will fail with an error if the plan file cannot be found, cannot be parsed, or if it contains mission items that are [not supported](#supported_mission_commands).
```
shakthi-prashanth-m: >The method will fail with an error if the plan file cannot be found, cannot be parsed, or if it contains mission items that are [not supported](#supported_mission_commands).

The method will fail with an error if plan file cannot be found, **or** it cannot be parsed, or if it contains mission items that are [not supported](#supported_mission_commands).
(Added *or* )

en/guide/missions.md
hamishwillee: No, it is correct (what you have suggested is also fine).

en/guide/missions.md
hamishwillee: Now this is an interesting question - and a possible bug in implementation.

If you check out [MissionItem](https://docs.dronecore.io/v/develop/en/api_reference/classdronecore_1_1_mission_item.html) there is no API to land or takeoff! (as an aside, on PX4 if the takeoff command is omitted from mission an MC vehicle will take off without it)

That is why I have this note in the guide:

> The Mission API does not (at time of writing) support "repeat", takeoff, return to land etc. DroneCore provides some omitted functionality through the Action API.

And if you look at the code we [create the missionitem](https://github.com/dronecore/DroneCore/blob/f7354e3c515dae4bd3a902d334ccb39ca42f8c10/plugins/mission/mission_impl.cpp#L972) but do nothing for the land/takeoff handling. I think that is a bug. 

Most mission created by Mission Planner will have a takeoff as their first item and land as the last item. IMO we should implement land/takeoff because they are reasonable things to want to declare, and it isn't sensible to reject import because they are missing.

@julianoes Got a better idea?

en/guide/missions.md
hamishwillee: My way is correct (so is yours, but I only change things if the alternative is "better").

hamishwillee:
> @hamishwillee, do you need full screenshot of the QGC plan for higher resolution ? I did crop the image just enough to cover mission items while uploading.

No. Ideally I like "full version" with QGC shrunk as far as is reasonable and map zoomed to fill screen -
 essentially then you get QGC just being a "frame" around the flight path. You can kind of get the idea from this: https://docs.dronecore.io/v/develop/en/examples/transition_vtol_fixed_wing.html

But the cropped version you provided is fine too! Certainly not worth your effort to regenerate at this point.

hamishwillee:
OK, so now this is just waiting on discussion of how mission import of takeoff/land should be implemented (https://github.com/dronecore/docs/pull/100#discussion_r168985430) and any associated docs changes. Perhaps we should just merge this now and carry on the discussion separately?

en/guide/missions.md
shakthi-prashanth-m: OK. [We've considered takeoff and land](https://github.com/dronecore/DroneCore/blob/develop/plugins/mission/mission_impl.cpp#L972-L973) while importing QGC mission. But however these commands are "looked" as waypoint mission items with lat, lon, alt parameters and I remember @julianoes mentioned once that mission items are sent to firmware and firmware will simply execute them as waypoint command. @julianoes, please confirm this.

However, I feel it is better to send mission items as "takeoff", "land" commands to firmware. Reason is that
- Even after land command, vehicle stays in "Mission" & "Armed" state unless application commands for "RTL".
- Firmware won't show "Landing detected" in PX4 logs.

julianoes: > @julianoes mentioned once that mission items are sent to firmware and firmware will simply execute them as waypoint command. @julianoes, please confirm this.

I think what I meant is that you don't need to add a takeoff waypoint because the firmware will automatically insert a takeoff waypoint, so DroneCore shouldn't worry about it.

> - Even after land command, vehicle stays in "Mission" & "Armed" state unless application commands for "RTL".

That sounds like a bug. At least if auto disarm is on it should disarm after landing.

> - Firmware won't show "Landing detected" in PX4 logs.

That would also be a bug.

However, I think adding an RTL or land command at the end of the mission is just not supported by DroneCore yet.

shakthi-prashanth-m: I tested with a QGC plan containing only takeoff and land as mission items. First I tried upload and fly using QGC. After mission accomplished, it left the vehicle in **Mission Disarmed** state. QGC Mission editor on the right showed command IDs: Takeoff, Land. Firmware shows **Landing detected** when it executes last mission item.
With DroneCore mission interface, a difference as we know is that QGC Mission editor shows both takeoff and land as Waypoint mission items. Firmware does **NOT** show **Landing detected** at the end.

shakthi-prashanth-m: >However, I think adding an RTL or land command at the end of the mission is just not supported by DroneCore yet.

Yes

hamishwillee: OK, so I've never though omitting takeoff and was sensible - if only because Plane/VTOL don't auto insert them like MC does, and a waypoint at "0" altitude is not a landing - it is "very low flight" :-). 

Upshot, I agree the problem is that we don't support them yet, so let's do so: https://github.com/dronecore/DroneCore/issues/286
I haven't bothered to create issue for missing  RTL because that seems very "optional" to me.

With respect to this particular issue, my "solution" is to add a note that these are "supported for import only": https://github.com/dronecore/docs/pull/100/commits/ae773b35da304513bb5ace825156d2b7ff4c98fd - I could go into more detail on what happens in detail for import, but lets just fix this going forward. 

Merging this now and the current docs accurately reflect the API.

PR #103, opened 2018-02-27
Update docs for new add_any_connection method

hamishwillee:
Adds docs for https://github.com/dronecore/DroneCore/pull/264

### DIFF
--- a/en/guide/connections.md
+++ b/en/guide/connections.md
@@ -6,17 +6,25 @@ In order to detect vehicles you must first specify the communication ports that 
 
 ## Monitoring a Port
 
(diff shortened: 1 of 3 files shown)


en/guide/connections.md
```diff
@@ -6,17 +6,25 @@ In order to detect vehicles you must first specify the communication ports that
...
 
 ```cpp
 DroneCore dc;
-DroneCore::ConnectionResult connection_result = dc.add_udp_connection();
+DroneCore::ConnectionResult connection_result = dc.add_any_connection(); 
 ASSERT_EQ(connection_result,DroneCore::ConnectionResult::SUCCESS)
 ```
 
-> **Tip** This is the normal/most common way to connect to PX4 over WiFi, using the standard PX4 broadcast UDP port for off-board APIs (14540). You can also explicitly specify the connection properties/ports if needed.
+> **Note** The default connection is to the standard PX4 broadcast UDP port for off-board APIs (14540). This is the normal/most common way to connect to PX4 over WiFi.
```
julianoes: Would be good to add: `This port is also used by MAVROS, however, QGroundControl uses port 14550`.

hamishwillee: I get your point, but I didn't want to confuse things by mentioning QGC as this is not an offboard API. Instead what I have done is linked to the [standard UDP port information](https://dev.px4.io/en/simulation/#default-px4-mavlink-udp-ports), which includes QGC.

PR #145, opened 2018-11-20
Fix up ActionResult to Action::Result

hamishwillee:
Doc fixes following on from https://github.com/Dronecode/DronecodeSDK/pull/595


julianoes:
Thanks a lot @hamishwillee!

PR #151, opened 2019-02-06
ApiRef: Add mission raw plugin

hamishwillee:
Note, I couldn't add passthrough plugin yet, because it does not get built.

### DIFF
--- a/en/cpp/README.md
+++ b/en/cpp/README.md
@@ -51,6 +51,13 @@ The most important classes are:
 - [LogFiles](/api_reference/classdronecode__sdk_1_1_log_files.md): Download log files from the vehicle.
 
 
+The following APIs provide more direct access to underlying MAVLink messages/types.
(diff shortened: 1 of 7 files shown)


en/cpp/README.md
```diff
@@ -51,6 +51,13 @@ The most important classes are:
 - [LogFiles](/api_reference/classdronecode__sdk_1_1_log_files.md): Download log files from the vehicle.
 
 
+The following APIs provide more direct access to underlying MAVLink messages/types.
```
hamishwillee: @julianoes Not in particular this bit - which separates out the direct APIs from the others. I don't know whether you wanted to say something else, but I figured worth explaining why there are multiple ways to access APIs.

julianoes: Ok!

julianoes:
Thx @hamishwillee

PR #155, opened 2019-03-21
Fix PX4 SITL command

julianoes:
With master and 1.9.x it is `px4_sitl` instead of `posix`.


julianoes:
Yer dependencies are hard. Well, my editor just removes trailing whitespace aggressively (and I think all editors should :smile: ).

julianoes:
Thanks.

PR #162, opened 2019-06-04
Update C++ docs following build system changes

JonasVautherin:
Update the C++ docs, now showing how to build using cmake.

@hamishwillee: I made the edits with gitbook-editor, and it somehow made some changes (like escaping parentheses `\(`). Should I revert that or is it fine to keep them?

@julianoes: I removed the build section about Docker, as I am not sure if it is up-to-date and I believe it is not so relevant (people who may want to do that would go check our scripts, and it is noise for people who don't know docker). I can revert it if you want.

### DIFF
--- a/en/contributing/build.md
+++ b/en/contributing/build.md
@@ -29,7 +29,7 @@ To build the *Dronecode SDK* C++ Library on Linux (or macOS after installing the
 1. First install the dependencies
    ```bash
    sudo apt-get update -y
-   sudo apt-get install cmake build-essential colordiff astyle git libcurl4-openssl-dev libtinyxml2-dev doxygen -y
+   sudo apt-get install cmake build-essential colordiff git doxygen -y
    ```
    > **Note** If the build reports a missing dependency, confirm that the set above matches the requirements in the [appropriate docker file for your platform](https://github.com/Dronecode/DronecodeSDK/tree/{{ book.github_branch }}/docker).
 
@@ -53,16 +53,15 @@ To build the *Dronecode SDK* C++ Library on Linux (or macOS after installing the
    ```
 1. Build the (debug) C++ library by calling:
    ```sh
-   make default
+   cmake -DCMAKE_BUILD_TYPE=Debug -DBUILD_SHARED_LIBS=ON -Bbuild/default -H.
+   cmake --build build/default
    ```
-   > **Tip** You can build *release* binaries by setting `BUILD_TYPE=Release` as shown below:
+   > **Tip** You can build *release* binaries by setting `CMAKE_BUILD_TYPE=Release` as shown below:
      ```sh
-     make clean
-     BUILD_TYPE=Release make
+     cmake -DCMAKE_BUILD_TYPE=Release -DBUILD_SHARED_LIBS=ON -Bbuild/default -H.
+     cmake --build build/default
      ```
-
-
-1. (Optionally) "Install" the *Dronecode SDK* [as described below](#install-artifacts). This is required in order to build [Dronecode SDK C++ apps](../guide/toolchain.md), but not to run SDK test code.
+1. "Install" the *Dronecode SDK* [as described below](#install-artifacts).
 
 
 ### Windows {#build_cpp_windows}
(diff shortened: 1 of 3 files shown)


en/contributing/build.md
```diff
@@ -94,31 +91,24 @@ To build the *Dronecode SDK* C++ Library on Windows:
...
-     cmake --build . --config Release
+     cmake --build build --config Release
      ```
    <span></span>
    > **Tip** To use more than one CPU core to build set this before building:
      ```sh
      set CL=/MP
      ```
 
-1. (Optionally) "Install" the SDK [as described below](#install-artifacts). This is required in order to build [Dronecode SDK C++ apps](../guide/toolchain.md), but not to run SDK test code.
```
hamishwillee: If this step is going to be optional then it is useful to explain why/when it is needed. So perhaps either add back a note like I had (or something better), or remove "Optionally".

en/contributing/test.md
```diff
@@ -1,90 +1,85 @@
...
+cd path/to/Firmware/
 make px4_sitl gazebo
 ```
 
 Then run the tests as shown:
-```
-make run_integration_tests
-```
-
-### Run With a Real Vehicle
```
hamishwillee: Is this deletion of running on a real vehicle section intentional?

If so, might be worth just adding the warning back at the end of the section:

> **Warning** Some of the tests might not be suited for real vehicles, especially the takeoff and kill test!

hamishwillee:
@JonasVautherin I have no way to evaluate whether they are "right" but the instructions certainly look cleaner. A few comments inline.

> I made the edits with gitbook-editor, and it somehow made some changes (like escaping parentheses \(). Should I revert that or is it fine to keep them?

Fine to keep them. They do no harm. Personally I always edit in a text editor, and will probably delete these again because they seem so very pointless :-). The are good sometimes for escaping words that contain underscores, which might otherwise be rendered in italic. 

> I removed the build section about Docker, as I am not sure if it is up-to-date and I believe it is not so relevant (people who may want to do that would go check our scripts, and it is noise for people who don't know docker). I can revert it if you want.

It is probably better to have no instructions than ones that don't work. That said, it is probably is worth mentioning that they exist and that people should the scripts if they follow the instructions here and the build fails.

en/contributing/build.md
JonasVautherin: It's like any library, in the end. You can install it on your system if you want to have it on your system, or you can install it locally if you want to have it installed locally. There is no particular reason for one or the other. For instance I'm on Arch, and I don't install stuff manually on my system: I leave that to my package manager.

Maybe I should remove this bullet point completely and add information on how to use it (with a sample CMakeLists.txt for instance).

en/contributing/test.md
JonasVautherin: To me this was adding noise. If somebody has absolutely no clue to the point where they may not realize that killing a real vehicle is dangerous, then probably we don't want to give them instructions on how to do that. And if they understand the difference between a simulator and a real vehicle, they probably have no issues running their code on both, do they?

Also we don't run them on real vehicles, so I would not consider that as "supported". I don't want to have to care about only adding "safe" integration tests just because somebody may blindly follow this documentation :sweat_smile:. What if somebody has a filter to not run the "dangerous" tests, and then we add a new kill test that doesn't get filtered and we crash the drone?

That's an opinion, I can bring it back if you want, that's your call!

hamishwillee: A lot of people are idiots, but then most of those don't bother to read the instructions. I would be tempted to add just a note at the end. It can omit specific tests. Something like:

> **Warning** The tests are designed to run in simulation, and may not be safe if run on a real vehicle.

JonasVautherin: I added the warning :+1:

en/contributing/build.md
JonasVautherin: I eventually went for removing the "Optionally". People who don't know what they are doing will be fine with installing, in the end :-).

JonasVautherin:
Regarding Docker, I would vote to put that in a "CI" page (under "Contributing"), that would explain how our CI works. Because the CI changes quite often, I would keep that high-level, explaining how it is organized and pointing to the scripts for people who are interested. Our docs is not the place where we will teach people about Jenkins/Docker/Travis anyway.

hamishwillee:
Thanks for fixes. This is all good from my point of view, so now merging. 

> Regarding Docker, I would vote to put that in a "CI" page (under "Contributing"), that would explain how our CI works. Because the CI changes quite often, I would keep that high-level, explaining how it is organized and pointing to the scripts for people who are interested. Our docs is not the place where we will teach people about Jenkins/Docker/Travis anyway.

Works for me.  leaving CI stuff for you to do as a separate task.

PR #167, opened 2019-06-30
Improve intro branding. Add status info

hamishwillee:
- Remove flight-stack specifics.
- Add python link to getting started. Clarify that it had/does everything.
- General changes to make it clear that C library doesn't need to be built on ubuntu/fedora


en/examples/README.md
```diff
@@ -54,8 +54,21 @@ See [QGroundControl > Download and Install](https://docs.qgroundcontrol.com/en/g
...
+
+Then build the example:
+```sh
+cd example/takeoff_land/
+mkdir build && cd build
+cmake ..
+make
+```
+
+#### macOS
```
hamishwillee: @julianoes Note, added macOS section, because I presume this cannot use rpm or deb files.

julianoes: Ok, we need to add support via brew soon.

PR #169, opened 2019-07-16
Improve instructions for iOS build

JonasVautherin:
@douglaswsilva: FYI

Updating the instructions to show how to build `mavsdk_server` for iOS.

### DIFF
--- a/en/contributing/build.md
+++ b/en/contributing/build.md
@@ -158,13 +158,9 @@ cmake --build build/default --target install
 
 ## Build for Android {#build_cpp_android}
 
-The MAVSDK programming-language-specific libraries share a common backend (called "mavsdk_server"), which may optionally be built as part of the C++ library. This section describes how to build it for Android, which is necessary only for contributors. If you want to use MAVSDK on Android, go to the [MAVSDK-Java](https://github.com/mavlink/MAVSDK-Java) documentation.
-
-Building for Android is a cross-compilation like the others, and can be done using dockcross (e.g. with the `dockcross/android-arm64` image). For that, go to the [cross-compilation](#cross_compilation_dockcross) section below.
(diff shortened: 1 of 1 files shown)


en/contributing/build.md
```diff
@@ -186,19 +186,19 @@ make android install
 
 ## Build for iOS {#build_cpp_iOS}
 
-> **Warning** You must first build the C++ Library (as shown above).
+> **Warning** You must first build the C++ Library with `mavsdk_server` (i.e. with `-DBACKEND=ON`, as shown [Building the Backend](#build_backend)).
```
hamishwillee: Just to confirm, to build for iOS you first build the backend (covered in that section below), then you build CPP library for iOS - that is what is covered here right? 
Then there is a further step not covered here for building swift or some other library to run witht he SDK - right?

1. It might be worth spelling this all out here rather than just forward linking to building the backend. Minimally you should explain why you must first build the server.
2. There is an inconsistency in terminology now - you refer to both mavsdk_server and the backend. One or the other. If it were me I would change the name of the section Building the Backend to Building the MAVSDK Server - you can choose to leave the link anchor the same if you want. 

I would also add a paragraph in that section explaining what the server does for you (ie after "The MAVSDK programming-language-specific libraries (e.g. Swift, Python) share a common backend, which may optionally be built as part of the C++ library.")

JonasVautherin: Got it! Will do! :blush:

JonasVautherin:
@hamishwillee: very sorry for the delay :see_no_evil:.

hamishwillee:
Hi @JonasVautherin,
No worries - thanks for that. I subedited a bit and this will need your recheck.

For the iOS bit I moved the mavsdk_server comment as a note at the end just stating that the build is needed for extending python/swift, but not for using them or C++. The point being that people will build the backend now by default but will then see that they can choose not to. 

Similarly, I removed this mavsdk_server comment from Android - at this point it makes no sense, and moved it into the cross compilations section.

JonasVautherin:
That's perfect, thanks a lot!

PR #176, opened 2019-12-02
Examples: Create into/links for all examples

hamishwillee:
This adds links and summary text for all examples into the example entry point page. I also alphabetized the examples.

I would have liked to create full docs for each, but this is a good start for anyone browsing.

This flows out of a recent query about how to use multiple drones.

@julianoes What's the difference between the two multiple drone examples? Feel free to edit the text directly and merge.

### DIFF
--- a/en/examples/README.md
+++ b/en/examples/README.md
@@ -6,12 +6,20 @@ This section contains examples showing how to use MAVSDK.
 
 Example | Description
 --- | ---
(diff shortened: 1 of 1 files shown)


julianoes:
Thanks @hamishwillee!

> What's the difference between the two multiple drone examples? Feel free to edit the text directly and merge.

I don't know, I need to check to merge them potentially.

hamishwillee:
Thanks. OK, I'll merge this for now, and we can update if needed.

PR #181, opened 2020-02-21
update access to information with pair pointer

Katawann:


hamishwillee:
Thanks @Katawann !

@julianoes This looks good to me - I'd not seen the `std::pair` before. Is there anything else you know we need to update either in docs or examples? ie. this appears to be used in params, Action and LogFile too.

julianoes:
I'd have to go through it all. Best case every example from docs is extracted from a real file which is tested in CI. Without that we will break things like this every time.

hamishwillee:
That would be nice :-(. Merged this one. Thanks @Katawann !

PR #182, opened 2020-02-26
add note to avoid Promise already satisfied

Katawann:
related to issue [https://github.com/mavlink/MAVSDK/issues/996](https://github.com/mavlink/MAVSDK/issues/996)

### DIFF
--- a/en/guide/telemetry.md
+++ b/en/guide/telemetry.md
@@ -189,6 +189,7 @@ Similarly, you can use the asynchronous method and block
     // Now ready to arm
 }
 ```
+> **Note** Make sure to unsubscribe from the callback after it suceed using `telemetry->health_all_ok_async(nullptr)`. Otherwise you will run into an error because `prom->set_value()` is called more than once.
 
 Depending on the architecture of your application, you may even wish to arm the vehicle in your callback function. 
 Usually though it is easier to understand program flow using the approach above.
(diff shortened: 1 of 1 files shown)


en/guide/telemetry.md
```diff
@@ -189,6 +189,7 @@ Similarly, you can use the asynchronous method and block
     // Now ready to arm
 }
 ```
+> **Note** Make sure to unsubscribe from the callback after it suceed using `telemetry->health_all_ok_async(nullptr)`. Otherwise you will run into an error because `prom->set_value()` is called more than once.
```
hamishwillee: spelling is "succeed".

hamishwillee: @JonasVautherin Anything to add to this?

Katawann: Don't need to add it in the code example IMO. The example is working while it is not called more than once

Katawann: I need to do a new commit to propose your new note ?

hamishwillee: @Katawann You would have but I did it for you now (I would have done this first time, but I wanted to make sure it still addresses your concern).

@JonasVautherin @julianoes Can either of you confirm that you're now happy with this PR?

hamishwillee:
Merged! Thanks very much @Katawann !

PR #183, opened 2020-03-02
getting_started: add install for Linux and macOS

julianoes:
We can now install the library using pre-built binaries on some of the  platforms.


julianoes:
I like your suggestions for installation but I would not necessarily add the quickstart because it would create duplication.

hamishwillee:
A quickstart is something different. It basically shows the simplest end to end process for getting up and running. Arguably it should be "per language". It is OK to duplicate a little :-)

hamishwillee:
I'm just trying to re-write this now as an installation guide only.

The other alternative is to split it into a quick start for each programming language.

hamishwillee:
@julianoes OK, so this is now "pure installation". I think this is a useful page. 

As I said though, we could make this a C++ quickstart instead. At the top we could link to quick starts for python and swift.

Then it would be:
- Installation
- Simulation (from https://mavsdk.mavlink.io/develop/en/examples/)
- Build example (from https://mavsdk.mavlink.io/develop/en/examples/)
- Next steps

Thoughts? If you're unconvinced, then I suggest we merge this and think on it.

en/getting_started/installation.md
```diff
@@ -0,0 +1,75 @@
...
+```
+brew install mavsdk
+```
+
+### Windows
+
+> **Note** No binaries for Windows are available yet.
+  For now you must [build the library from source](../contributing/build.md).
+
+### Fedora
```
julianoes: I would keep all Linux distributions together, so I'd just move Ubuntu down to here.

en/getting_started/installation.md
```diff
@@ -0,0 +1,75 @@
...
+## Swift (iOS) {#ios}
+
+[MAVSDK-Swift on GitHub](https://github.com/mavlink/MAVSDK-Swift) as well as the [reference docs](http://dronecode-sdk-swift.s3.eu-central-1.amazonaws.com/docs/master/index.html) should contain everything needed for iOS development of MAVSDK apps.
+
+
+## Python
+
+Use `pip3` to install [MAVSDK-Python](https://github.com/mavlink/MAVSDK-Python#mavsdk-python):
+```bash
+pip3 install mavsdk
```
julianoes: We need to add a note here to say that this currently only works for Windows, macOS, and Linux x86_64 but not for Linux arm platfroms like RPi yet.
See: https://github.com/mavlink/MAVSDK-Python/issues/117

en/getting_started/installation.md
```diff
@@ -0,0 +1,78 @@
...
+```
+
+### Development
+
+For development, or unsupported platforms, you can [build the library from source](../contributing/build.md).
+
+
+## Swift (iOS) {#ios}
+
+[MAVSDK-Swift on GitHub](https://github.com/mavlink/MAVSDK-Swift) as well as the [reference docs](http://dronecode-sdk-swift.s3.eu-central-1.amazonaws.com/docs/master/index.html) should contain everything needed for iOS development of MAVSDK apps.
```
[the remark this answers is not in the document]
julianoes: That will change soon, once Swift PM works for iOS, so don't bother.

julianoes:
> That implies that if we use this path of installation we may need to install additional dependencies when 

That's a good point! The dependencies should be taken care of the package management system. So for instance, with `apt-get`/`dpkg`/`yum` it should tell you "bla is missing" because the dependencies are listed. However, I just checked that and the dependencies might be missing. I will follow up with that and fix it.

For Arch Linux the dependencies will be installed by `yay`, they are listed here:
https://aur.archlinux.org/packages/mavsdk/

For macOS everything is compiled statically, so it should just work.

hamishwillee:
Actually @julianoes I wasn't even thinking about those dependencies :-0. I was thinking more Cpp compiler and cmake, which developers will need to _use_ our libraries.

julianoes:
> I was thinking more Cpp compiler and cmake, which developers will need to use our libraries.

Oh, that's here already: https://mavsdk.mavlink.io/develop/en/contributing/build.html
`build_essentials` and `cmake`.

hamishwillee:
Do they get included in the release binaries? Awesome.

julianoes:
@hamishwillee no they are only build time dependencies but not run time dependencies, so not needed.

hamishwillee:
Hi @julianoes 

OK, so I prototyped a quickstart here: https://github.com/mavlink/MAVSDK-docs/blob/pr_quickstart/en/getting_started/quickstart_cpp.md

This shows how to install library on all platforms, but only how to build examples on macos and linux. The reason for this is that for most other variants I do not know how to set up the simulator or  what dependencies are needed. For Windows I know, but that's not a quick installation process, and in fact it is a bit "manual". 

At the top you can see links to imaginary quickstarts for Python and Swift.

One thing this made me realise is that despite our discussions on Friday we don't HAVE to show how to install the build toolchain. They can use what they like though we use the one that is used to build the flight stack (here I say "get it when you set up the dev env). 

So that means we don't need a whole quick start to cover the setup as I thought

So, we have choices. If you like the quickstart, then I can evolve/integrate it a little more. If you're not sure I propose we merge the Installation version here. There are benefits to both approaches, and I honestly can't say which is better. I have a slight leaning towards keeping the current version here, but adding a note about the build toolchain that we expect - e.g:

> **Note** You can use whatever build toolchain you like. Note however that this project uses cmake and the same compiler used to compile the flight stack software (generally you will install these when setting up the simulator).

julianoes:
@hamishwillee thanks!

So I think we want this page just describing the install anyway, right?

And then the quickstart is good but of course it would be better if the "installation" of PX4 for just SITL purposes would be easier.

hamishwillee:
@julianoes I updated this with both an installation (general) and C++ quickstart. I'm feeling pretty smug about it too :-). 

Can you please sanity check this.

Note, there is a little more follow up work we might reasonable do - specifically point the app building doc to installation and installing the required toolchain

julianoes:
Great, thanks @hamishwillee!

PR #199, opened 2020-05-12
contributing: use documented -S instead of -H

julianoes:
`-H` is an undocumented cmake flag that we should not rely upon.

See: https://stackoverflow.com/a/54279862/8548472

### DIFF
--- a/en/contributing/build.md
+++ b/en/contributing/build.md
@@ -52,12 +52,12 @@ To build the MAVSDK C++ Library on Linux (or macOS after installing the [precond
    ```
 1. Build the (debug) C++ library by calling:
    ```sh
-   cmake -DCMAKE_BUILD_TYPE=Debug -DBUILD_SHARED_LIBS=ON -Bbuild/default -H.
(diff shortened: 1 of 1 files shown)


JonasVautherin:
See my comment in https://github.com/mavlink/MAVSDK/pull/1090. I used `-H` where `-S` was failing, not sure if that is solved everywhere. `-S` will work everywhere at some point, but it seems to me that `-H` is legacy, and in some cases needed (because it's still much nicer than `mkdir build && cd build && cmake` IMO).

julianoes:
Ok, so the issue might be with cmake 3.5.1 on Ubuntu 16.04.

So in that case it might be correct to use `-H` everywhere until we drop support for 16.04.

TSC21:
> Ok, so the issue might be with cmake 3.5.1 on Ubuntu 16.04.
> 
> So in that case it might be correct to use `-H` everywhere until we drop support for 16.04.

Btw I am using Ubuntu 18.04 and I can't seem to use -S.

julianoes:
Let's leave it as is with `-H`.

PR #203, opened 2020-05-30
en: add links to new Python reference docs

julianoes:
We can try to integrate them later, for now we have something.


hamishwillee:
This is good, but I'd be chasing Lorenz to see if you can get a better docs URL than http://mavsdk-python-docs.s3-website.eu-central-1.amazonaws.com/

PR #205, opened 2020-06-03
Improve contributing docs

julianoes:
For more info, see individual commit.

@JonasVautherin it would be great if you could fill out the TODO about different proto structures.


julianoes:
Thanks @hamishwillee for reviewing and improving!

julianoes:
@JonasVautherin I've filled in the TODO. If you could review it that would be good.

PR #216, opened 2020-08-10
Add Python QuickStart

JonasVautherin:


en/python/quickstart.md
```diff
@@ -0,0 +1,100 @@
...
+```
+commander takeoff
+commander land
+```
+
+The simulated drone should takeoff and land. If it doesn't, it may mean that SITL is not ready, or that there is a problem.
+
+## Take off from MAVSDK
+
+When we know that the simulator is ready, we can open an `apython` REPL:
```
hamishwillee, line 37: What does this mean? Open a file? i.e. what am I seeing here?

hamishwillee: Ignore, answered before.

hamishwillee:
Love it. I have subedited and merged. Check it out - fixes can be a post process.

julianoes (commented):
Nice, thanks @JonasVautherin!

PR #221, opened 2020-10-10
api_changes: describe API changes with v0.33.0

julianoes:
Goes with https://github.com/mavlink/MAVSDK/pull/694/.

### DIFF
--- a/en/cpp/api_changes.md
+++ b/en/cpp/api_changes.md
@@ -23,14 +23,19 @@ This means that breaking changes to the API result in a bump of the major versio
 
 ## v0.33.0
 
-There are some changes in how systems are discovered, connected and accessed with this release. The previous APIs are still available but marked deprecated and they will be removed in the future. The deprecations are also implemented as compiler warnings.
+There are some changes in how systems are discovered, connected and accessed with this release.
+The previous APIs are still available but marked deprecated and they will be removed in the future.
+The deprecations are also implemented as compiler warnings.
 
 ### Discovery of systems
 
-The 64bit (8 bytes) uuid is deprecated because it needed to be extended to 18 bytes. For more info, see this [MAVLink pull request](https://github.com/mavlink/mavlink/pull/786). Therefore, the discovery no longer contains the uuid. If the uid is required it can be found in the [Info plugin](../api_reference/classmavsdk_1_1_info.md#classmavsdk_1_1_info_1a812ed66265b7427bc781faec3f0fa89e).
+The 64bit (8 bytes) uuid is deprecated because it needed to be extended to 18 bytes.
+For more info, see this [MAVLink pull request](https://github.com/mavlink/mavlink/pull/786).
+Therefore, the discovery no longer contains the uuid.
+If the uid is required it can be found in the [Info plugin](../api_reference/classmavsdk_1_1_info.md#classmavsdk_1_1_info_1a812ed66265b7427bc781faec3f0fa89e).
 
 **Previous way of discovering systems:**
-```
+```cpp
 std::promise<void> discover_promise;
 auto discover_future = discover_promise.get_future();
 
@@ -43,7 +48,7 @@ discover_future.wait();
 ```
 
 **New way of discovering a system:**
-```
+```cpp
 std::promise<void> discover_promise;
 auto discover_future = discover_promise.get_future();
 
@@ -60,13 +65,18 @@ discover_future.wait();
 
 ### Accessing systems
 
-We aim to make the API of MAVSDK as easy and safe to use as possible. However, as we'll see below, this still comes with trade-offs.
+We aim to make the API of MAVSDK as easy and safe to use as possible.
+However, as we'll see below, this still comes with trade-offs.
 
-When designing the API we decided to expose the MAVSDK user a reference to a system `System&` which means the system will always be instantiated and there won't be any segfaults trying to access a system that is null.
+When designing the API we decided to expose the MAVSDK user a reference to a system `System&`, which means the system will always be instantiated and there won't be any segfaults trying to access a system that is null.
 This mostly worked fine, however, had some drawbacks:
-- Without an actual system discovered the `System&` is basically a null reference that won't really work. This is not really obvious unless you check its system and component IDs for 0. This is not very intuitive and also means additional checks inside the plugins are required.
-- The ownership model is not very clear as `Mavsdk` owns the `System` but then each plugin only seems to require the `System`. Internally, the plugins still require `Mavsdk` to be alive and working but that's not apparent through the API.
-- If there are multiple sytems they need to be accessed using the `uuid`. This seems like a crutch because essentially we would just like to have a `std::vector` of systems but of course that's not posible with references.
+- Without an actual system discovered the `System&` is basically a null reference that won't really work.
+  This is not really obvious unless you check its system and component IDs for 0.
+  This is not very intuitive and also means additional checks inside the plugins are required.
+- The ownership model is not very clear as `Mavsdk` owns the `System` but then each plugin only seems to require the `System`.
+  Internally, the plugins still require `Mavsdk` to be alive and working but that's not apparent through the API.
+- If there are multiple sytems they need to be accessed using the `uuid`.
+  This seems like a crutch because essentially we would just like to have a `std::vector` of systems but of course that's not posible with references.
 
 As the API for the uid was about the change anyway (see above) it seemed time to redesign the API to access systems with these considerations in mind.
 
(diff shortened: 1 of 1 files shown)


JonasVautherin (approved):
Nice!

en/cpp/api_changes.md
```diff
@@ -21,9 +21,91 @@ This means that breaking changes to the API result in a bump of the major versio
...
+There are some changes in how systems are discovered, connected and accessed with this release.
+The previous APIs are still available but marked deprecated and they will be removed in the future.
+The deprecations are also implemented as compiler warnings.
+
+### Discovery of systems
+
+The 64bit (8 bytes) uuid is deprecated because it needed to be extended to 18 bytes.
+For more info, see this [MAVLink pull request](https://github.com/mavlink/mavlink/pull/786).
+Therefore, the discovery no longer contains the uuid.
+If the uid is required it can be found in the [Info plugin](../api_reference/classmavsdk_1_1_info.md#classmavsdk_1_1_info_1a812ed66265b7427bc781faec3f0fa89e).
```
hamishwillee, line 35: Is this link correct? ie have you updated the api ref. Couldn't see the UUID at the link

en/cpp/api_changes.md
```diff
@@ -21,9 +21,91 @@ This means that breaking changes to the API result in a bump of the major versio
...
+discover_future.wait();
+```
+
+**New way of discovering a system:**
+```cpp
+std::promise<void> discover_promise;
+auto discover_future = discover_promise.get_future();
+
+mavsdk.subscribe_on_change([&mavsdk, &discover_promise]() {
+    const auto system = mavsdk.systems().at(0);
```
hamishwillee, line 56: The `.at(0)` is a bit confusing for two reasons:
1. I think this is because systems is now a shared point to a system, and you're grabbing the first system. But at this point you don't know about that - perhaps a forward reference to next section.
2. What if you're grabbing the second system? Does the user have any way to grab the system that actually got added?

hamishwillee:
Modulo the comment looks great. The corresponding changes need to be made in the body of the docs as well

Minimally
- e.g. update to example https://mavsdk.mavlink.io/develop/en/examples/takeoff_and_land.html
- https://mavsdk.mavlink.io/develop/en/guide/connections.html

But every single doc shows how to get a system, and they might need to be modified.

en/cpp/api_changes.md
julianoes: It's supposed to go here: https://mavsdk.mavlink.io/develop/en/api_reference/classmavsdk_1_1_info.html#classmavsdk_1_1_info_1a812ed66265b7427bc781faec3f0fa89e

And then if you click on the actual struct https://mavsdk.mavlink.io/develop/en/api_reference/structmavsdk_1_1_info_1_1_identification.html there is a hardware_uid.

en/cpp/api_changes.md
julianoes: 1. Oh, you're right. Maybe a comment like `// see how to access systems below`?

julianoes:
@hamishwillee thanks for the review! I'll go through the code examples and change them as well.

en/cpp/api_changes.md
JonasVautherin: 1. I don't get that, sorry :see_no_evil:. `mavsdk.systems()` gives you the list of systems, and because you know it just changed, you can be fairly sure that there is one system and that it is connected. I don't know if it is a feature or not, but I believe that we never remove any `System` from the `systems()` list, so the first one discovered will always be at [0] (it's not like the list will get shuffled or something).
2. I believe that whether or not you get an indication of which system changed will not matter in your implementation in this function. Right now, anyway `subscribe_on_change` is used both for discovery and for checking the connection state, so you will have this logic anyway. I could imagine separating them by having another callback as part of a `System`, though :thinking::

```c++
mavsdk.subscribe_on_discover(<callback>); // This tells me that there is a new system in the `systems()` list
system.subscribe_on_change(<callback>); // This tells me that this particular system's state has changed
```

Would that make sense? Otherwise `systems()` as we have it now is also fine with me, but I don't think that the id brings value :+1:.

julianoes: 1. This is only a doc thing. We are using the new API but it gets explained further below in the docs.
2. `mavsdk.subscribe_on_discover` so this would trigger every time that a new system is added, and they will never be removed because, ever.
And then I would maybe call it `system_subscribe_on_connection_change` but then you still don't know if it has just connected or disconnected.

JonasVautherin: > but then you still don't know if it has just connected or disconnected

Right. If the only state that can change is `is_connected()`, then it could also be something like `system.subscribe_is_connected([](bool is_connected) { ... });` And if we think that there could be other things in the state, then `subscribe_on_connection_change` tells you that you may want to check `system.is_connected()`, which seems fine to me :+1:.

A bonus from splitting them is that then we don't care about whether or not the `systems()` list changes (so later we could even remove systems if we wanted to): `subscribe_on_discover` tells you that you will find a new system in the list, and if you are interested you can get a pointer to it, but after that you will never need to access this system through that list again (because you just keep the pointer).

julianoes: `system.subscribe_is_connected([](bool is_connected) { ... });` makes sense, perfect.

So I don't understand your pointer idea exactly. Can you put that in code?

JonasVautherin:
```c++
// Discovery step
auto discover_prom = std::promise<std::shared_ptr<System>>{};
auto discover_fut = discover_prom.get_future();
mavsdk.subscribe_on_discover([&mavsdk, &prom]() {
    prom.set_value(mavsdk.systems().at(0));
    mavsdk.subscribe_on_discover(nullptr);
});

auto system = discover_fut.get();

// From this point on, I will never need to call `mavsdk.systems()` again,
// as I will just work with my `system`. So I don't care if the `systems()`
// list gets reshuffled from time to time.
system->subscribe_is_connected([](bool is_connected) {
    if (!is_connected) {
        std::cout << "ERROR: system got disconnected, which is not supported by this script" << std::endl;
        exit(1);
    }
});

// Create my plugins and use them
Action action(system);
...
```

en/cpp/api_changes.md
hamishwillee: Cool. Though it refers if you follow the links to uid2 of mavlink, and I have no idea what that is?

julianoes:
@JonasVautherin updated as discussed.

JonasVautherin (approved):

en/cpp/api_changes.md
```diff
@@ -21,9 +21,116 @@ This means that breaking changes to the API result in a bump of the major versio
...
+- Nice getter `std::vector<std::shared_ptr<System>> systems()` allowing easier access to all systems.
+
+**Previous way to get a system:**
+```cpp
+System& system = mavsdk.system();
+```
+
+**New way to get a system:**
+```cpp
+std::shared_ptr<System> system = mavsdk.systems().at(0);
```
dakejahl, line 120: Should this be getting the system by uuid? Could use a list instead of a vector. 
```suggestion
std::shared_ptr<System> system = mavsdk.systems().at(uuid);
```

JonasVautherin: I don't have a strong opinion on that, except that I never used it this way :see_no_evil:. `mavsdk.systems()` is the interface to the discovered systems. So unless you hardcoded the `uuid` somewhere in your script, at this point you don't know what it will be. Typically, to know it you'll have to go through all the systems (usually there aren't so many) and check them individually.

And there, I have personally never had a use-case where I check the uuid. I do check if it has an autopilot (`system->has_autopilot()`), which tells me that it is a drone. When it isn't, I assume it is a ground station.

How do you use the uuid?

dakejahl: >So unless you hardcoded the uuid somewhere in your script, at this point you don't know what it will be

This is exactly what I am doing :joy: 

I don't have strong opinions either way, it just feels more intuitive to grab a system by ID rather than by index. If you have more mavlink components than a single FC and a single GCS, you are going to be manually choosing `system_id`s anyway. Choosing by index would mean you have to select it first and then check the ID, or you must know in which order they enumerate.

JonasVautherin: > it just feels more intuitive to grab a system by ID rather than by index

Oh right. I would agree with that. But the point here is that you just get a list of systems, and if you expect more than one you should go through them and do your own logic (whether it uses the uuid or flags like `has_autopilot`). Here the `systems().at(0)` is a way to do `systems().giveMeTheOnlySystemThere()` :laughing:.

dakejahl: yeah this makes sense, keeps the API minimal

PR #224, opened 2020-11-27
Pass after API changes

julianoes:
I had some time on the plane and went through the examples and guides to update them to the latest APIs and usage recommendations. I've removed the copied example sources because - in my opinion - they will always be-out-of-date and wrong, so it's better to link to them.


en/guide/README.md
```diff
@@ -1,5 +1,5 @@
 # C++ API Guide
 
-This API guide provides topics showing how to use the SDK APIs to perform common (and not so common) tasks. 
+This API guide provides topics showing how to use the MAVSDK APIs to perform common (and not so common) tasks.
```
hamishwillee, line 3: FWIW I think using "SDK" is fine in this context.

en/guide/telemetry.md
```diff
@@ -182,43 +184,43 @@ This same approach can be used to report only messages that meet some condition.
...
     });
-    
+
     future_result.get(); //Block until promise is fulfilled.
     // Now ready to arm
 }
 ```
 > **Note** Note that `prom->set_value()` must only be called once.
-  If your code can potentially call this multiple times, remember to unsubscribe from the callback after it succeeds (e.g. using `telemetry->health_all_ok_async(nullptr)`).
+  If your code can potentially call this multiple times, remember to unsubscribe from the callback after it succeeds (e.g. using `telemetry.subscribe_all_ok_async(nullptr)`).
```
hamishwillee, line 223: Just checking, is it really `subscribe_all_ok_async` - ie not having "health" seems mismatch with `subscribe_health_all_ok`

hamishwillee:
Looks good to me on "scan", but I'm not an expert :-). A few comments above.

> I've removed the copied example sources because - in my opinion - they will always be-out-of-date and wrong, so it's better to link to them.

Well as you have to maintain them, fair enough - you've linked to the branch specific build/version, so all good!
FYI I liked them as they provided a copy that I could DIFF against when I updated (to see what else I might have missed). Also because I could see a lot more hits of particular API usage on search (effectively including example code in the search engine)

en/guide/telemetry.md
julianoes: Oops it's `subscribe_health_all_ok`, thanks for finding that.

hamishwillee: No worries. I presume you'll tidy as a post-process.

julianoes: Done https://github.com/mavlink/MAVSDK-docs/pull/225.

PR #226, opened 2021-01-17
Remove the warning for linux arm support of mavsdk-python

ytaskiran:
Since Linux ARM platforms are [supported ](https://pypi.org/project/mavsdk/#files) now, there is no need to raise this note for MAVSDK-Python installation.


JonasVautherin (approved):
Thanks!

PR #237, opened 2021-03-29
api_changes: add note about QGC plan import change

julianoes:
To import QGC missions, you now need to use MissionRaw instead of Mission.

Thanks @mcsauder!

### DIFF
--- a/en/cpp/api_changes.md
+++ b/en/cpp/api_changes.md
@@ -20,8 +20,28 @@ This means that breaking changes to the API result in a bump of the major versio
 > **Note** At time of writing, breaking/incompatible changes *are not* resulting in the major version number being increased!
   We plan to release version 1.0.0 in the near future, after which the above strategy will be adopted (i.e. currently breaking changes can occur in both minor and patch releases).
 
+## 0.38
 
-## v0.33.0
+### Mission
+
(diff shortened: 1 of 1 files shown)


hamishwillee:
@julianoes This is good, and I have merged it. 

Note though that I might expect some more background within the MissionRaw docs - for example, I know that previously the idea was that you'd import a consistent mission into SDK objects, that could then be manipulated in Mission and then reuploaded. Further, if the object in mission not supported by MAVSDK you might reject the import. It isn't clear to me how this now works from this information provided - my assumption is that now you can download anything and just save it as a buffer to file - but we know nothing about the format of the data or how it is used -(I'm not familiar with MissionRaw, perhaps this is obvious).

julianoes:
> I know that previously the idea was that you'd import a consistent mission into SDK objects, that could then be manipulated in Mission and then reuploaded.

Yes you could. Initially it would reject a mission during import if it contained mission items that were not supported (had no mapping to MAVSDK mission items). Then this was extended to just warn on unsupported items. This can make for a frustrating experience if you upload your mission were some of the actions are ignored but you don't check/see the emitted warning. The drone would then e.g. not trigger the camera, move the servo, whatever it actually should be doing.

Now, with mission raw it's just a one to one mapping.

hamishwillee:
@julianoes Thanks for the explanation. I looked at the MissionRaw, and yes, it makes complete sense.

PR #240, opened 2021-05-27
new logging callback feature documented

levevg:


en/cpp/guide/logging.md
```diff
@@ -0,0 +1,46 @@
+# Logging
+MAVSDK core and plugins output some useful log messages during their work.
+By default, the messages are printed to `stdout`. You may want to override this
+behavior, e.g. redirect messages to an external logging system, or disable
+printing messages unless they are important enough.
+
+## Usage
+It is possible to customize logging with a user-defined callback function:
+```c++
+#include "mavsdk/log_callback.h"
```
julianoes, line 10:
```suggestion
#include <mavsdk/log_callback.h>
```

en/cpp/guide/logging.md
```diff
@@ -0,0 +1,46 @@
...
+
+// ...
+
+mavsdk::log::subscribe([](mavsdk::log::Level level,   // message severity level
+                          const std::string &message, // message text
+                          const std::string &file,    // source file from which the message was sent
+                          int line) {                 // line number in the source file
```
julianoes, line 17:
```suggestion
mavsdk::log::subscribe([](mavsdk::log::Level level,   // message severity level
                          const std::string& message, // message text
                          const std::string& file,    // source file from which the message was sent
                          int line) {                 // line number in the source file
```

en/cpp/guide/logging.md
```diff
@@ -0,0 +1,46 @@
...
+```
+
+## Important notes
+Please note that callback can be called from different threads. Synchronization
+is not handled on MAVSDK's side for performance reasons. You may want to
+implement synchronization yourself if your logging system isn't thread
+safe already.
+
+Avoid doing long duration heavy tasks in the callback, it should return as
+quickly as possible.
```
julianoes, line 46: Good points!

julianoes (commented):
Awesome, that is great! Thanks a lot. I only commented some minor things that I would prefer different, but overal this is spot on!

PR #242, opened 2021-07-22
Improve connection docs, and update with main

julianoes:
See commit messages for more detail.


en/cpp/api_reference/classmavsdk_1_1_param_server.md
```diff
@@ -0,0 +1,253 @@
+# mavsdk::ParamServer Class Reference
+`#include: param_server.h`
+
+----
+
+
+Provide raw access to retrieve and provide server parameters.
```
hamishwillee, line 7: What is a server parameter? I know what a MAVLink parameter is, but server parameter is a new term for me.

en/cpp/guide/connections.md
```diff
@@ -1,39 +1,100 @@
...
+if (connection_result != ConnectionResult::Success) {
+    std::cout << "Adding connection failed: " << connection_result << '\n';
+    return;
+}
+```
+
+
+### Connecting over UDP
+
+When connecting over UDP, there are two setups to distinguish: server and client mode.
```
hamishwillee, line 48: So essentially the normal mode we use for SDK to act like a server (GCS) and listen. We can ONLY do this for the local host right? You can't specify that you want to listen on some other interface?

hamishwillee:
I added a couple of comments/questions mostly FMI. But it looks really good so merging.

en/cpp/api_reference/classmavsdk_1_1_param_server.md
julianoes: It's a Server that provides MAVLink parameters. Does that make sense?

en/cpp/guide/connections.md
julianoes: Correct.

en/cpp/api_reference/classmavsdk_1_1_param_server.md
hamishwillee: Not really. A mavlink component or system has parameters. I have never seen a mavlink server flying around :-). 

So let me guess without bothering to read anything. Is this a server on the internet that somehow knows about parameters of one or more mavlink systems? If not, then there is probably some kind of naming problem.

julianoes: Let me try to rephrase. So there are a couple of mavlink microservices, as we call them. And most of them have some sort of client-server architecture. For instance, there is the ground station uploading a mission, and then there is an autopilot answering to that request. So in this case the ground station is the mission client and the autopilot is the mission server. It serves to the mission request.

In the same way we have a client like MAVSDK or QGC querying parameters and an autopilot providing/serving them.

So the plugin which has been added here is the autopilot side. Basically, all `xxx_server` plugins are on the vehicle side answering to the requests. Basically, these server plugins allow you to bridge your own custom autopilot software to mavlink, on the vehicle side, or to implement a camera onboard which can talk mavlink parameters.

Should we improve the doc string to make that clearer?

hamishwillee: I think you're saying this plugin runs on the vehicle. That doesn't sound right. Perhaps we could have a brief conversation about this?

julianoes: Yes, e.g. on the companion side. Sure...

hamishwillee: @julianoes Ah, that was not obvious to me. Is there a **guide** document that covers this use case? If so, it should cross link to your server plugins so that someone can see this is what is going on from either direction.

julianoes: There is not yet but that's a good idea. I'll put it on my list.

PR #243, opened 2021-09-07
Add server plugins

julianoes:
This updates the API reference and adds a new page to describe the server plugins.
@bazfp feel free to extend that page.


hamishwillee (approved):
Looks awesome!!

I modified the server-plugins page a bit; I wanted to explain why they are called server plugins, which is not done by explaining that MAVLink defines protocols using client-server pattern and the server is usually vehicle side.

Anyway, I will merged, but you might disagree with the change, in which case, feel free to change to your version.

julianoes:
@hamishwillee yes that's better, thank you!

PR #244, opened 2021-09-07
Add why C++ and gRPC to FAQ

julianoes:
I'm writing this down as @bazfp had asked me.

### DIFF
--- a/en/faq.md
+++ b/en/faq.md
@@ -94,3 +94,31 @@ From a MAVSDK project point of view, there is also an advantage of not having a 
 The requirements for MAVSDK were that it can efficiently run in embedded setups, e.g. as part of an app or onboard a drone on the companion computer which suggested C or C++.
 The goal with MAVSDK was to provide a simple and safe API. This favored C++ over C as it allows for more expressive but type-safe APIs using standard library containers (e.g. `std::vector<MissionItem>`).
 
+### Why is gRPC used for the language wrappers?
+
+There are multiple ways to support multiple programming languages all with their pros and cons:
+
+1. Separate implementation per language:
+   + Nice native API.
+   + No non-native dependencies (often) in language wrappers.
+   - Lots of implementation effort required, does not scale well with a small team.
+   - Languages will not have consistency of feature parity.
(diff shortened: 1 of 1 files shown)


JonasVautherin (approved):
Nice! If I may just comment on this:

> we underestimated just how much work was required to write all the auto-generation for the various function signatures (sync, async, result handling, streams, etc.), and also were not aware about the pain that comes with every dependency.

I agree on the dependencies side, but I think that the auto-generation would be required for the language bindings as well. And using `protoc` with the proto IDL is not too bad, I would say (if the alternative is to write our own IDL and parser and all...).

I'm all for moving to direct bindings, but I would definitely do that through the autogen. Just remove protobuf and gRPC and use only `protoc-gen-mavsdk` with the proto IDL :+1:

julianoes:
@JonasVautherin ok, do you want to change the wording slightly then? Feel free to make suggestions.

JonasVautherin:
No no I just wanted to share more details, but what you wrote is perfectly fine IMO :+1: 

I would merge it as-is but the CI is failing for some reason :innocent:

en/faq.md
```diff
@@ -88,3 +88,37 @@ While it would be nice to have access to all MAVLink messages in the language wr
...
+
+The requirements for MAVSDK were that it can efficiently run in embedded setups, e.g. as part of an app or onboard a drone on the companion computer which suggested C or C++.
+The goal with MAVSDK was to provide a simple and safe API. This favored C++ over C as it allows for more expressive but type-safe APIs using standard library containers (e.g. `std::vector<MissionItem>`).
+
+### Why is gRPC used for the language wrappers?
+
+There are multiple ways to support multiple programming languages all with their pros and cons:
+
+1. Separate implementation per language:
+   + Nice native API.
```
hamishwillee, line 102: The + and - will all render as bullets. Either you need to make them code style or prefix them with a bullet.

hamishwillee:
Looks good other than the "bullet" issue above. I'll look at the CI issue. I think it is a problem with gitbook rotting unsupported and somehow getting out of sync with plugins.

PR #251, opened 2021-11-04
Fix server plugin page links

dlech:
Links were using `..` instead of `.` causing them to point to the wrong relative directory.


julianoes:
They seem to be broken indeed. Thanks!

PR #260, opened 2022-05-11
Update plugins page to be up to date

junwoo091400:
## Document that gets updated
https://mavsdk.mavlink.io/main/en/cpp/contributing/plugins.html

## Changes
- Updated Folder diagram to keep it up to date with MAVSDK repository (as plugins moved into src/mavsdk)
- Update *Note* part since in the doc the 'Note' part gets replaced with an image, to make it read more natural

### DIFF
--- a/en/cpp/contributing/plugins.md
+++ b/en/cpp/contributing/plugins.md
@@ -1,6 +1,6 @@
 # Writing Plugins
 
-MAVSDK-C++ is split into a [core](https://github.com/mavlink/MAVSDK/tree/{{ book.github_branch }}/src/core) and multiple independent [plugins](https://github.com/mavlink/MAVSDK/tree/{{ book.github_branch }}/src/plugins).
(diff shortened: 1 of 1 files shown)


en/cpp/contributing/plugins.md
```diff
@@ -119,7 +122,7 @@ message SetReturnToLaunchAltitudeResponse {
 }
 ```
 
-> **Note** that requests can defined SYNC, ASYNC, or BOTH using `option (mavsdk.options.async_type) = ...;`.
```
junwoo091400, line 122: @hamishwillee Can this be replaced with "::: note" block like PX4 User Doc? If so, I think it would be better to use that system :thinking:

julianoes (approved):
Oh, that's great, thanks for that @junwoo091400!

julianoes:
Sorry, took a while to merge this :smile: .

PR #262, opened 2022-05-22
cpp: how to swap MAVLink headers and dialects

julianoes:
This adds a section how the MAVLink headers can be replaced, and how a custom dialect can be used.


hamishwillee (approved):
It is worth saying that common.xml must be present/included by some dialect?
I'm going to merge this because it is excellent as is.

julianoes:
Thanks for the fixup. Is common.xml required? :thinking: Probably, since MAVSDK would need it but if you use your own dialec, I think, you're aware of that.

PR #267, opened 2022-07-21
Run codespell on docs

hamishwillee:
This is the result of running codespell on the docs like:
```
codespell -s -L "ned"
```
Other than a deep desire to change `PositionNed ==> positioned` nothing found in protos. 
Docs had a few small typos, all fixed.

From which I can only assume that there is spell checking going on already :-)


hamishwillee:
FYI @julianoes

julianoes (commented):
All the separate variants! That's funny. Thanks @hamishwillee!

PR #268, opened 2022-11-07
Update telemetry.md; telemetry.subscribe_all_ok() does not exist

raceking37:
Found a small error in the documentation: telemetry.subscribe_all_ok() doesn't exist, this was probably meant to be the telemetry.subscribe_health_all_ok() method.

### DIFF
--- a/en/cpp/guide/telemetry.md
+++ b/en/cpp/guide/telemetry.md
@@ -220,7 +220,7 @@ Similarly, you can use the asynchronous method and block
 }
 ```
 > **Note** Note that `prom->set_value()` must only be called once.
-  If your code can potentially call this multiple times, remember to unsubscribe from the callback after it succeeds (e.g. using `telemetry.subscribe_all_ok(nullptr)`).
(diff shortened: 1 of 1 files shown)


hamishwillee (approved):
Thanks! 
@julianoes I'm merging because there doesn't seem to be anything else this can be. More generally, should we be saying what health OK means? I assume it means OK to arm "as far as we know" but it might mean "OK to arm canonically" or something else.

julianoes:
Thanks, makes sense.

There is the flag `health.is_armable()` which tells you whether arming is possible:
https://github.com/mavlink/MAVSDK/blob/849cb9a3c3bd18d21731da3b0c4fe31ca6b9d039/src/mavsdk/plugins/telemetry/include/plugins/telemetry/telemetry.h#L410

hamishwillee:
So what does healthy "mean" then in MAVlink terms? Yes I could look it up. I guess I'm hinting we need to be a bit more clear about how the data can be used in the docs.

hamishwillee:
Perhaps replace it with something more meaningful in next revision then?

julianoes:
Perhaps yes.

PR #278, opened 2024-01-03
Make label on front of version v2 say v2 - not main

hamishwillee:
As it is now

image

### DIFF
--- a/book.json
+++ b/book.json
@@ -3,7 +3,7 @@
     "title": "MAVSDK Guide",
     "variables": {
         "logo": "./assets/site/sdk_logo_full.png",
-        "github_branch": "main"
+        "github_branch": "v2.0"
     },
     "plugins": [
         "youtube",


julianoes (approved):
Thanks!

hamishwillee:
Cool. We'll look at moving to Vitepress in the new year.

julianoes:
I'd also like to include the Python docs: http://mavsdk-python-docs.s3-website.eu-central-1.amazonaws.com/

hamishwillee:
You mean include them as part of the single docs package? Python docs are generated with doxygen or something else? 

We need to think about this. The hand crafted C docs parser I created is a little flakey. We might be better off looking at using doxygen for generating all reference docs. If we want to import within Vitepress to have a central home, we can possibly then do that by importing the HTML. It looks fairly flexible so this might well be possible. Needs more investigation.

julianoes:
Sure, we need to figure out how to generate the files for it.

hamishwillee:
For Python maybe https://niklasrosenstein.github.io/pydoc-markdown/ - I used this for generating the libmav python docs from some pretend Python classes (see https://github.com/Auterion/libmav-python/blob/main/docs/build/docs/content/libmav.md).
What I like about this is that it supports many input/output formats so you can probably get it to work quite well on the sources you already have.

But that still leaves C++. My parser is not robust enough, so we need a better way to get markdown for that.

PR #279, opened 2024-03-08
Update connections.md

dakejahl:
For local client connections the url string is slightly different. The only place I saw that this was documented was in the source.


en/cpp/guide/connections.md
```diff
@@ -85,6 +85,17 @@ if (connection_result != ConnectionResult::Success) {
 }
 ```
 
+E.g. to connect to a local mavlink-router client endpoint at 14551, you would do:
```
julianoes, line 88: What about?

```suggestion
E.g. to listen locally to a mavlink-router endpoint at 14551, you would do:

```

julianoes (commented):
Thanks. I would probably call it listen to rather than connect to, right?

julianoes (approved):

PR #285, opened 2024-07-29
Remove plugin library linking instructions from docs

KarthiAru:
The following sections were removed from the 'taking_off_landing.md' and 'telemetry.md' documentation files:

1. Instructions for linking the 'mavsdk_telemetry' plugin library to the application's CMake build definition file.
2. Instructions for linking the 'mavsdk_action' plugin library to the application's CMake build definition file.

With recent versions of MAVSDK, you only need MAVSDK::mavsdk and hence these sections are removed.


KarthiAru:
Creating a new PR with more related changes.

